PR #2, opened 2019-10-15
Feature/order code

StefanGajic:


fetcher/book.go
```diff
@@ -0,0 +1,65 @@
...
+type identifier struct {
+	ISBN10      []string `json:"isbn_10"`
+	ISBN13      []string `json:"isbn_13"`
+	Openlibrary []string `json:"openlibrary"`
+}
+type Author struct {
+	Name string `json:"string"`
+}
+
+type Cover struct {
```
sfqi: You can unexport these two structs, I don't think they're used outside the package.

fetcher/book.go
```diff
@@ -0,0 +1,65 @@
...
+	key := fmt.Sprintf("ISBN:%v", isbn)
+	rawBook, ok := result[key]
+	if !ok {
+		return nil, errors.New("Value for given key cannot be found")
+	}
+	var book Book
+	sliceOfBytes, err := rawBook.MarshalJSON()
+	if err != nil {
+		fmt.Println("Error converting *rawMessage into []byte")
+	}
```
sfqi: I missed this earlier:
1. Gotta return the error
2. Why marshal a `json.RawMessage` object? Can't it just me converted to a byte slice? We might not even need to convert it to a byte slice, try to `Unmarshal` it directly.
(also I'm not too happy with `sliceofBytes`, a var name should tell us what's in the slice - but if 2. is right, we don't need a var anyway)

fetcher/book.go
```diff
@@ -0,0 +1,65 @@
...
+
+	err = json.Unmarshal(sliceOfBytes, &book)
```
sfqi: Error handling!

fetcher/book.go
```diff
@@ -0,0 +1,65 @@
...
+	fmt.Println(book.Identifier.ISBN10[0])
+	fmt.Println(book.Identifier.ISBN13)
```
sfqi: Let's remove these two.

main.go
```diff
@@ -60,73 +59,34 @@ func GetBooks(w http.ResponseWriter, r *http.Request) {
...
+	ISBN string `json:"ISBN"`
+}
 
-func fetchBooks(isbn string) (*Book, error) {
-	url := basePath + isbn + queryParams
-	fmt.Println(url)
-	response, err := http.Get(url)
-	if err != nil {
-		return nil, err
+func WriteBooks(w http.ResponseWriter, r *http.Request) {
```
sfqi: How many books is this endpoint writing? Just one, no?
Also, let's use `Create` instead of `Write` - this is supposed to be the C from CRUD, and you've got the `createBookRequest` struct already.

main.go
```diff
@@ -5,11 +5,10 @@ import (
 	"fmt"
 	"net/http"
 
-	"github.com/pkg/errors"
-)
+	f "./fetcher"
```
sfqi: Don't use relative paths. Also no need for the `f`.

fetcher/book.go
```diff
@@ -0,0 +1,65 @@
+package fetcher
```
sfqi: Got a better idea for the package name, let's go with `openlibrary`, so when you call a function, it will be `openlibrary.FetchBook`

fetcher/book.go
```diff
@@ -0,0 +1,65 @@
...
+func FetchBook(isbn string) (*Book, error) {
+	url := basePath + isbn + queryParams
+	fmt.Println(url)
+	response, err := http.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	status := response.StatusCode
+	fmt.Println("Status code: ", status)
+	defer response.Body.Close()
```
sfqi: Move this up, right after the error check - good practice.

fetcher/book.go
bokimilinkovic: yes it can be done with:
err = json.Unmarshal([]byte(*rawBook), &book)

bokimilinkovic:
In function GetBooks, slice books( 'books := []BookModel' ) should be global

sfqi:
> In function GetBooks, slice books( 'books := []BookModel' ) should be global

@bokimilinkovic Didn't even notice that. But you do it your upcoming PR.

PR #9, opened 2019-10-17
Fixed error handling

StefanGajic:
https://trello.com/c/FAjgno30/24-fix-error-handling

### DIFF
--- a/openlibrary/book.go
+++ b/openlibrary/book.go
@@ -2,7 +2,6 @@ package openlibrary
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"net/http"
 )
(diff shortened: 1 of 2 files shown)


openlibrary/book.go
```diff
@@ -44,18 +44,24 @@ func FetchBook(isbn string) (*Book, error) {
...
+		err := fmt.Errorf("error while decoding from FetchBook: %v", err)
 		return nil, err
 	}
 
 	key := fmt.Sprintf("ISBN:%v", isbn)
+	fmt.Println(key)
 	rawBook, ok := result[key]
 	if !ok {
-		return nil, errors.New("Value for given key cannot be found")
+		errorKey := fmt.Sprintf("value for given key cannot be found: %s", result[key])
```
sfqi: How about `fmt.Errorf`, like above?

openlibrary/book.go
```diff
@@ -44,18 +44,24 @@ func FetchBook(isbn string) (*Book, error) {
 	result := make(map[string]*json.RawMessage, 0)
 	err = json.NewDecoder(response.Body).Decode(&result)
 	if err != nil {
+		err := fmt.Errorf("error while decoding from FetchBook: %v", err)
 		return nil, err
 	}
 
 	key := fmt.Sprintf("ISBN:%v", isbn)
+	fmt.Println(key)
```
sfqi: We don't need to print this.
And also, for simple string concatenation: `key := "ISBN:" + isbn`

PR #14, opened 2019-10-22
Testing fetchBook finished

bokimilinkovic:


openlibrary/book_test.go
```diff
@@ -48,4 +53,52 @@ func TestFetchBook(t *testing.T) {
...
+		if responseBook != nil {
+			t.Errorf("Expected to be nil but got: %v", responseBook)
+		}
+
+	})
+
+	t.Run("Book with given ISBN cannot be found", func(t *testing.T){
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte(`{"ISBN:0140447938": {"title": "War and Peace (Penguin Classics)"}}`))
```
sfqi: Can you make the ISBN value more different from the one passed to `FetchBook`? Took me a while to figure out what's going on (the test name being unclear didn't help).

Also, isn't this scenario already [covered](https://github.com/sfqi/library/blob/master/openlibrary/book_test.go#L31)? Although the test name seems wrong in that one as well.

main.go
```diff
@@ -12,7 +13,7 @@ import (
 )
 
 type BookModel struct {
-	Id            int    `json:"id"`
+	Id            int    `json:"id,omitempty"`
```
sfqi: Well this should never be empty. I guess we should add `id` initialization to `CreateBook`. Don't do it now, though, just remove the `omitempty`

main.go
```diff
@@ -58,7 +59,11 @@ var books = []BookModel{
 
 func GetBooks(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
-	json.NewEncoder(w).Encode(books)
+	err := json.NewEncoder(w).Encode(books)
+	if err != nil {
+		http.Error(w,"Not found",http.StatusNotFound)
```
sfqi: This should be a bad request (something's wrong with the json request body).

And another problem: now we don't have "logging" (we were using the `fmt` to log errors), we just return them in the response - both are needed.

main.go
```diff
@@ -70,18 +75,18 @@ func CreateBook(w http.ResponseWriter, r *http.Request) {
...
 	if err := json.NewDecoder(r.Body).Decode(&createBook); err != nil {
-		fmt.Println("error while decoding in CreateBook: ", err)
+		http.Error(w,"Error while decoding from request body",400)
 		return
 	}
 	fmt.Println(createBook.ISBN)
 	book, err := client.FetchBook(createBook.ISBN)
 	if err != nil {
-		fmt.Println("error while fething book: ", err)
+		http.Error(w,"Error while fetching book: " + err.Error(),400)
```
sfqi: Use `http.StatusBadRequest`. Although this should be internal server error - if the open library server is down, it doesn't have anything to do with the request.

We can improve the openlibrary client later, so that it differentiates between different kinds of errors, i.e. so that it can tell us that there's no book with that ISBN - will make a task for it.

Also: see above about logging.

main.go
```diff
@@ -70,18 +75,18 @@ func CreateBook(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
 
 	if err := json.NewEncoder(w).Encode(book); err != nil {
-		fmt.Println("error while encode from CreateBook: ", err)
+		http.Error(w,"Internal server error:"+err.Error(),500)
```
sfqi: We don't have logging.
Use http constants.

main.go
```diff
@@ -90,22 +95,32 @@ func UpdateBook(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 	var book BookModel
 	err := json.NewDecoder(r.Body).Decode(&book)
+	if strconv.Itoa(book.Id) != "0" {
```
sfqi: Nice thought, but there are better ways of doing that (we'll get there soon).

main.go
```diff
@@ -90,22 +95,32 @@ func UpdateBook(w http.ResponseWriter, r *http.Request) {
...
+		errors.New("Id of the book can not be changed")
```
sfqi: This doesn't really do anything. But you'll remove this entire block anyway (see above)

main.go
```diff
@@ -90,22 +95,32 @@ func UpdateBook(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
-		fmt.Println("error while decoding from UpdateBook", err)
+		http.Error(w,"Error while decoding from request body",400)
 		return
 	}
 	params := mux.Vars(r)
 	id, err := strconv.Atoi(params["id"])
 	if err != nil {
-		fmt.Println("error converting string to int from UpdateBook", err)
+		http.Error(w,"Error while converting url parameter into integer",400)
```
sfqi: Logging, constant.

main.go
```diff
@@ -90,22 +95,32 @@ func UpdateBook(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	if id > len(books){
+		// if we have 3 books, and pass 5 as id, its out of range... should be bad request?
+		http.Error(w,"Given index is out of bounds",400)
 		return
 	}
```
sfqi: `len` isn't the best way to do this, the slice index isn't the same as the `Book.Id`. Move this inside the loop

main.go
```diff
@@ -120,7 +135,11 @@ func GetBook(w http.ResponseWriter, r *http.Request) {
 	params := mux.Vars(r)
 	id, err := strconv.Atoi(params["id"])
 	if err != nil {
-		fmt.Println("error converting string to int from GetOneBook", err)
+		http.Error(w,"Error while converting url parameter into integer",400)
```
sfqi: Logging, constant

main.go
```diff
@@ -120,7 +135,11 @@ func GetBook(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	if id >len(books){
+		http.Error(w,"Given index is out of bounds",400)
```
sfqi: Same as in update. Also, I was lazy and didn't notice that `GetBook` doesn't use the same mechanism to find a book as `UpdateBook` - change it (maybe it could be a database method called `FindById`).

main.go
```diff
@@ -129,7 +148,11 @@ func GetBook(w http.ResponseWriter, r *http.Request) {
 			book = books[i]
 		}
 	}
-	json.NewEncoder(w).Encode(book)
+	err = json.NewEncoder(w).Encode(book)
+	if err != nil {
+		http.Error(w,"Error encoding response into book", 500)
```
sfqi: Logging, constant

main_test.go
```diff
@@ -29,3 +33,24 @@ func TestGetBooks(t *testing.T) {
...
+
+	if err != nil {
+		t.Errorf("Error occured, %s",err)
+	}
+
+	rr := httptest.NewRecorder()
+	r:=mux.NewRouter()
+
+	r.HandleFunc("/books/{id}",UpdateBook).Methods("PUT")
+	r.ServeHTTP(rr,req)
```
sfqi: This isn't wrong, but I'd rather try not to initialize a router, but to set the id param in the request. Check [this](https://github.com/gorilla/mux/pull/322/files) out.

Hit me up if there are any issues.

README.md
```diff
@@ -5,4 +5,6 @@ For setting env var in .bashrc to be permanent , open .bashrc file using:
 $sudo nano ~/.bashrc
 at the and of file, enter this line at the end of file:
 export LIBRARY='https://openlibrary.org/api/'
-Logout, and login again 
\ No newline at end of file
+Logout, and login again 
+
+We need to remove '/' at the end from https://openlibrary.org/api/ , and add it in bookPath
```
sfqi: We should just check the trailing `/` when initializing the openlibrary client. I'll make a card.

main_test.go
bokimilinkovic: how can i get response without router, and serveHTTP? Using http.Client.do(request) ?

main_test.go
```diff
@@ -29,3 +34,118 @@ func TestGetBooks(t *testing.T) {
...
+		if status := rr.Code; status != http.StatusOK {
+			t.Errorf("Status code differs. Expected %d. Got %d", http.StatusOK, status)
+		}
+		expected := `{"id":0,"title":"another title","author":"another author","isbn_10":"","isbn_13":"","OpenLibraryId":"another some id","cover":"","publish_date":""}` + "\n"
+		assert.Equal(t, expected, rr.Body.String(), "Response body differs")
+	})
+	t.Run("Error decoding Book attributes", func(t *testing.T) {
+		req, err := http.NewRequest("PUT", "/books/{id}", bytes.NewBuffer([]byte(`{"id":"12","title":zdravo}`)))
+		mapOfParams := map[string]string{}
+		mapOfParams["id"] = "2"
```
sfqi: I think you can do this in one command. Also just call it params.

openlibrary/book_test.go
```diff
@@ -19,33 +20,84 @@ func TestFetchBook(t *testing.T) {
...
+			t.Errorf("Got error: %s", err)
 		}
 
 		if responseBook.Title != "Concrete mathematics" {
-			t.Errorf("We did not get the expected response,expected %s, but got %s", expected, responseBook.Title)
+			t.Errorf("We did not get the expected response,expected %s, got %s", expected, responseBook.Title)
 		}
 	})
 
 	t.Run("book with error in title", func(t *testing.T) {
```
sfqi: This test doesn't really make sense, you can just remove it.

openlibrary/book_test.go
```diff
@@ -19,33 +20,84 @@ func TestFetchBook(t *testing.T) {
...
+		responseBook , err := client.FetchBook("0140447932111xxxx")
+		chechIfErrorIsNIl(err,t)
+		if err.Error() != "value for given key cannot be found: ISBN:0140447932111xxxx" {
+			t.Errorf("Got error : %s",err)
 		}
+		checkIfBookIsNil(responseBook,t)
 	})
 }
+
+func checkIfBookIsNil(b *Book,t *testing.T){
```
sfqi: Don't think we need this one

openlibrary/book_test.go
```diff
@@ -19,33 +20,84 @@ func TestFetchBook(t *testing.T) {
...
+	if b != nil{
+		t.Errorf("Expected Book to be nil, got : %v",b)
+	}
+}
+
+func chechIfErrorIsNIl(err error,t *testing.T){
```
sfqi: My idea was to have both the nil check **and** the error equality check in the func. 
```
if err == nil {
...
}
if err.Error != errorMsg {
...
}
```

Could we do that?

Also, typo: check, not chech.

main.go
```diff
@@ -144,6 +163,27 @@ func main() {
 	r.HandleFunc("/books", GetBooks).Methods("GET")
 	r.HandleFunc("/books", CreateBook).Methods("POST")
 	r.HandleFunc("/books/{id}", UpdateBook).Methods("PUT")
-	r.HandleFunc("/book/{id}", GetBook).Methods("GET")
+	r.HandleFunc("/book/{id}", FindBookById).Methods("GET")
```
sfqi: I said **database method** `FindById`. :)

main_test.go
```diff
@@ -29,3 +34,156 @@ func TestGetBooks(t *testing.T) {
...
+
+		rr := httptest.NewRecorder()
+		handler := http.HandlerFunc(UpdateBook)
+		handler.ServeHTTP(rr, req)
+		expectedError := "Book with given Id can not be found"
+		if status := rr.Code; status != http.StatusBadRequest && rr.Body.String() != expectedError {
+			t.Errorf("Expected status code: %d and error: %s,  got: %d and %s", http.StatusBadRequest, expectedError, status, rr.Body.String())
+		}
+	})
+	// Ostalo je jos testiranje za error Encoding, ne znam da li je to neophodno
```
StefanGajic: Why comment here in Serbian? Please remove it

main.go
```diff
@@ -91,46 +97,59 @@ func UpdateBook(w http.ResponseWriter, r *http.Request) {
...
 		}
 	}
+	if !found {
+		errorFindingBook(w,err)
+		return
+	}
 }
 
-func GetBook(w http.ResponseWriter, r *http.Request) {
+func FindBookById(w http.ResponseWriter, r *http.Request) {
```
StefanGajic: Why did you changed the name of function?

main_test.go
bokimilinkovic: dont cry

PR #20, opened 2019-10-28
moved BookModel to domain/model folder and renamed to Book

StefanGajic:
https://trello.com/c/4hSBfTCz/36-create-book-domain-model


sfqi:
Are the tests still passing?

PR #21, opened 2019-10-29
Put database to mock pkg

StefanGajic:
https://trello.com/c/IeyueehJ/40-take-the-database-out-of-handler-pkg

### DIFF
--- a/handler/handlers.go
+++ b/handler/handlers.go
@@ -20,33 +20,10 @@ import (
 var openLibraryURL = os.Getenv("LIBRARY")
 var client = *openlibrary.NewClient(openLibraryURL)
 
-type db struct {
-	id    int
-	books []model.Book
-}
-
-func (bm db) FindBookById(id int) (book model.Book, location int, found bool) {
-	for i, b := range bm.books {
-		if b.Id == id {
-			book = b
-			location = i
-			found = true
-			break
-		}
-	}
-
-	return book, location, found
-}
-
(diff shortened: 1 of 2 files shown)


handler/handlers.go
```diff
@@ -37,42 +38,9 @@ func (bm db) FindBookById(id int) (book model.Book, location int, found bool) {
...
-		Author:        "another author",
-		Isbn:          "another isbn",
-		Isbn13:        "another isbon13",
-		OpenLibraryId: "another some id",
-		CoverId:       "another cover ID",
-		Year:          "2019",
-	},
-}
-
 var shelf = &db{
```
sfqi: Isn't this the database?

repository/mock/database.go
```diff
@@ -0,0 +1,36 @@
+package mock
```
sfqi: We'll possibly end up renaming this, but meh, let's leave it for now.

repository/mock/database.go
```diff
@@ -34,3 +34,26 @@ var Books = []model.Book{
...
+	Id    int
+	Books []model.Book
+}
+
+var Shelf = &Db{
+	Id:    len(Books),
+	Books: Books,
+}
+
+func (bm Db) FindBookById(id int) (book model.Book, location int, found bool) {
```
bokimilinkovic: method over struct, should go with pointer (bm *Db)

PR #25, opened 2019-10-30
Feature/db operations

StefanGajic:
https://trello.com/c/ZhxGVpXa/43-implement-database-operations-findbyid-getall-create-update-as-mockdb-methods

### DIFF
--- a/handler/handlers.go
+++ b/handler/handlers.go
@@ -59,7 +59,7 @@ func (b BookHandler) CreateBook(w http.ResponseWriter, r *http.Request) {
 
 	bookToAdd := b.createBookModelFromBook(*book)
 
-	if err := b.db.Create(bookToAdd); err != nil {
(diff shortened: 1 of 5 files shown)


handler/handlers.go
```diff
@@ -35,69 +46,72 @@ type createBookRequest struct {
...
 	book, err := client.FetchBook(createBook.ISBN)
 	if err != nil {
 		fmt.Println("error while fetching book: ", err)
 		http.Error(w, "Error while fetching book: "+err.Error(), http.StatusInternalServerError)
 		return
 	}
 
-	bookToAdd := CreateBookModelFromBook(*book)
-	mock.Shelf.Books = append(mock.Shelf.Books, bookToAdd)
+	bookToAdd := b.createBookModelFromBook(*book)
```
bokimilinkovic: why is this method over handler, when we only call this function inside CreateBook?

PR #26, opened 2019-11-04
Encode dto.BookResponse instead of model.Book

StefanGajic:
https://trello.com/c/CRHp2E0z/48-encode-dtobookresponse-instead-of-modelbook-as-http-response

### DIFF
--- a/handler/handlers.go
+++ b/handler/handlers.go
@@ -12,6 +12,8 @@ import (
 
 	"github.com/library/domain/model"
 	"github.com/library/handler/dto"
+	openlibrary "github.com/library/openlibrary/dto"
+
 	"github.com/library/repository/mock"
 )
 
@@ -21,7 +23,7 @@ type BookHandler struct {
 }
 
 type openLibraryClient interface {
-	FetchBook(isbn string) (*dto.Book, error)
+	FetchBook(isbn string) (*openlibrary.Book, error)
 }
 
 func NewBookHandler(db *mock.DB) *BookHandler {
@@ -75,7 +77,7 @@ func (b *BookHandler) Create(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func (b *BookHandler) toBook(book *dto.Book) (bm *model.Book) {
+func (b *BookHandler) toBook(book *openlibrary.Book) (bm *model.Book) {
 
 	isbn10 := ""
 	if book.Identifier.ISBN10 != nil {
--- a/handler/handlers.go
+++ b/handler/handlers.go
@@ -30,7 +30,7 @@ func NewBookHandler(db *mock.DB) *BookHandler {
 	}
 }
 
-func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) {
+func (b *BookHandler) Get(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Content-Type", "application/json")
 	allBooks := b.Db.GetAllBooks()
@@ -126,8 +126,8 @@ func (b *BookHandler) Update(w http.ResponseWriter, r *http.Request) {
 		errorConvertingId(w, err)
 		return
 	}
-	book := &model.Book{}
-	book.Id = id
+	book := &dto.BookResponse{}
+	book.ID = id
 	book.Title = updateBookRequest.Title
 	book.Year = updateBookRequest.Year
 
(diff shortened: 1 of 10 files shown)


handler/handlers.go
```diff
@@ -12,6 +12,8 @@ import (
 
 	"github.com/library/domain/model"
 	"github.com/library/handler/dto"
+	openlibrary "github.com/library/openlibrary/dto"
```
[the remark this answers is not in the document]
sfqi: Also this PR seems to be doing two different things, adding openlibrary/dto & fixing the responses. Which, fine, but also I'd rather see it as two PRs.

handler/handlers.go
```diff
@@ -143,7 +145,7 @@ func (b *BookHandler) Update(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func (b *BookHandler) Get(w http.ResponseWriter, r *http.Request) {
+func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) {
```
sfqi: This seems to be the third thing done by the PR.
Let's fix this in a separate one.

handler/handlers.go
```diff
@@ -126,8 +128,8 @@ func (b *BookHandler) Update(w http.ResponseWriter, r *http.Request) {
 		errorConvertingId(w, err)
 		return
 	}
-	book := &model.Book{}
-	book.Id = id
+	book := &dto.BookResponse{}
```
sfqi: Leave the model one.

sfqi: The point was that in every handler method that has a `json.NewEncoder(w).Encode(book)` call, instead of passing a `*model.Book` to `Encode`, we should pass a `*dto.BookResponse`. So before encoding, a `model.Book` should be transformed into `dto.BookResponse`.

openlibrary/client.go
```diff
@@ -6,7 +6,7 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/library/handler/dto"
+	openlibrary "github.com/library/openlibrary/dto"
```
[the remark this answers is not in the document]
sfqi: Are we not removing this?

handler/handlers.go
```diff
@@ -158,7 +200,9 @@ func (b *BookHandler) Get(w http.ResponseWriter, r *http.Request) {
 		errorFindingBook(w, err)
 		return
 	}
-	err = json.NewEncoder(w).Encode(book)
+	bookResponse := toBookResponse(*book)
```
sfqi: I'd prefer it if `bookResponse` expected a pointer to `model.Book`.

handler/handlers.go
```diff
@@ -30,12 +32,44 @@ func NewBookHandler(db *mock.DB) *BookHandler {
...
+		response := dto.BookResponse{
+			ID:            book.Id,
+			Title:         book.Title,
+			Author:        book.Author,
+			Isbn:          book.Isbn,
+			Isbn13:        book.Isbn13,
+			OpenLibraryId: book.OpenLibraryId,
+			CoverId:       book.CoverId,
+			Year:          book.Year,
+		}
```
sfqi: Could you use the `toBookResponse` function here?

handler/handlers_test.go
```diff
@@ -27,31 +26,7 @@ func init() {
...
-	}
-
-	expected := `[{"id":1,"title":"some title","author":"some author","isbn_10":"some isbn","isbn_13":"some isbon13","olid":"again some id","cover":"some cover ID","publish_date":"2019"},{"id":2,"title":"other title","author":"other author","isbn_10":"other isbn","isbn_13":"other isbon13","olid":"other some id","cover":"other cover ID","publish_date":"2019"},{"id":3,"title":"another title","author":"another author","isbn_10":"another isbn","isbn_13":"another isbon13","olid":"another some id","cover":"another cover ID","publish_date":"2019"}]` + "\n"
-	fmt.Println(rr.Body.String())
-	fmt.Println(expected)
-	if rr.Body.String() != expected {
-		t.Errorf("handler returned unexpected body: got%v want %v",
-			rr.Body.String(), expected)
-	}
-}
```
sfqi: Wait, you just deleted the book index test here.

But okay, I wanted to make a card specifying that we should decode the response in handler tests, since defining the expected value should be easier that way. So we'll deal with this in another task.

Meanwhile, instead of deleting this, comment it out.

PR #27, opened 2019-11-04
Initial postgres setup

bokimilinkovic:


repository/postgres/postgres.go
```diff
@@ -0,0 +1,53 @@
...
+		panic(err)
+		return nil,err
+	}
+	store.Db = db
+	return &store,nil
+
+}
+func(db *dbStore)FindById(id int)(*Book, error){
+	b := Book{}
+	if err := db.Db.Where("id=?",id).First(&b).Error;err!=nil{
```
sfqi: Try `db.First(&b)`

repository/postgres/postgres.go
```diff
@@ -0,0 +1,53 @@
...
+	Title string
+	Year string
+}
+
+type Store interface {
+	FindById(id int)(*Book, error)
+}
+
+type dbStore struct {
+	Db *gorm.DB
```
sfqi: `DB`

repository/postgres/postgres.go
```diff
@@ -0,0 +1,53 @@
...
+	_ "github.com/lib/pq"
+
+
+)
+const(
+	host ="localhost"
+	port =5432
+	user="bojan"
+	password="bojan"
+	dbname="library"
```
sfqi: So I should create a `bojan` user on my machine if I want to run this app?
hint: openlibrary, url

repository/postgres/postgres.go
```diff
@@ -0,0 +1,53 @@
...
+)
+//this struct is just used for demonstration
+type Book struct{
+	gorm.Model
+	Title string
+	Year string
+}
```
sfqi: I mean, yeah, we could do this and separate the domain model from the repo model (maybe it's technically cleaner), but it might be kinda tedious. Let's just use the domain one for now.

main.go
```diff
@@ -23,6 +23,7 @@ func main() {
 		Olc: olc,
 	}
 
+
```
sfqi: ?

repository/postgres/postgres.go
```diff
@@ -0,0 +1,53 @@
+package postgres
+
+
+import(
+	"fmt"
+	"github.com/jinzhu/gorm"
+	_ "github.com/jinzhu/gorm/dialects/postgres"
+	_ "github.com/lib/pq"
```
sfqi: Do we need both of those? Genuinely asking.

repository/postgres/postgres.go
```diff
@@ -0,0 +1,53 @@
...
+//this struct is just used for demonstration
+type Book struct{
+	gorm.Model
+	Title string
+	Year string
+}
+
+type Store interface {
+	FindById(id int)(*Book, error)
+}
```
sfqi: Where do we use this?

bokimilinkovic: I created interface where i define all methods which we are going to use over our database ... I thought we need interface because i assume we will later be mocking and testing database.. If we dont need this i will delete

sfqi: That makes sense. But:
1. I want to try defining the interface somewhere else
2. Since it's not used anywhere at this point, let's wait til it's needed

repository/postgres/postgres.go
sfqi: Do you think `postgres` is a reliable password to be used in production? :)
(whispers: _make it all configurable_)

repository/postgres/postgres.go
bokimilinkovic: but dont we also need function Where() , to find exact given id?

repository/postgres/postgres.go
bokimilinkovic: you are right, we dont need github.com/lib/pq ... its included inside jinzhu/gorm , but we need dialects because postgres driver

main.go
```diff
@@ -15,14 +17,23 @@ func main() {
...
-	r := mux.NewRouter()
+	dbStore,err := postgres.Open()
+	if err != nil{
+		panic(err)
+	}
+	fmt.Println("Successfully connected to dataabse")
+
+	//just checking if this method is working fine
+	u,err:=dbStore.FindById(1)
+	fmt.Println(u)
```
sfqi: I mean we're not gonna merge this into master.

From trello:
For the moment, it will be easier to try it out outside od the library project: just import postgres into another play project. You can create a book record directly in the database in order to test the FindByID method.

repository/postgres/postgres.go
sfqi: Okay, I'll try to be clearer.
1. This way, _some_ things are configurable, not _all_ of them. Are we sure that the db will invariably run on that port?
2. There could a type, called `Config` per example, that would contain every attribute needed to open a connection. And it could be passed to `Open` as an argument.

repository/postgres/postgres.go
```diff
@@ -0,0 +1,44 @@
...
+	"os"
+)
+const(
+	host ="localhost"
+	port =5432
+	dbname="library"
+)
+
+var username = os.Getenv("PSQLUSERNAME")
+var password = os.Getenv("PSQLPASSWORD")
```
sfqi: If env var are being accessed from non-main packages, it might be time for some rethinking.

repository/postgres/postgres.go
```diff
@@ -0,0 +1,44 @@
...
+
+type dbStore struct {
+	DB *gorm.DB
+}
+
+func Open()(*dbStore, error){
```
sfqi: This should either have a connection string or a `Config` object as an argument.

The string (or the configuration) should be initialized by the caller.

As I said:
For the moment, it will be easier to try it out outside od the library project: just import postgres into another play project. You can create a book record directly in the database in order to test the FindByID method.

repository/postgres/postgres.go
```diff
@@ -0,0 +1,44 @@
...
+	store := dbStore{}
+	psqlInfo := fmt.Sprintf("host=%s port=%d user=%s password=%s dbname=%s sslmode=disable",
+		host,port,username,password,dbname)
+	db, err := gorm.Open("postgres",psqlInfo)
+	if err != nil {
+		panic(err)
+		return nil,err
+	}
+	b := model.Book{}
+	db.AutoMigrate(&b)
```
sfqi: We're using golang-migrate for migrations.

repository/postgres/postgres.go
```diff
@@ -0,0 +1,44 @@
...
+	store.DB = db
+	return &store,nil
+
+}
+func(db *dbStore)FindById(id int)(*model.Book, error){
+	b := model.Book{}
+	if err := db.DB.Where("id=?",id).First(&b).Error;err!=nil{
```
sfqi: Meh, sorry, `db.First(&b, id)`.

repository/postgres/postgres.go
```diff
@@ -5,39 +5,55 @@ import (
...
+	if err != nil{
+		panic(err)
+	}
+	var cfg Config
+	err = yaml.NewDecoder(f).Decode(&cfg)
+	if err != nil {
+		return nil,err
+	}
+	return &cfg,nil
+}
```
sfqi: I think it's cleaner if this kind of stuff is done by the caller (basically, the `main` pkg).
Imagine if we had more than five pkgs that require configuration in the way you've implemented, so they're all using yaml. And then the devops team says we're switching to env vars. Now you'd have to modify all those packages.
Or let's say someone is in charge of assembling all the configuration values. They would have to go through every package to check how and whether that pkg loads its configuration.
Or let's say you want to write a pkg that's used in multiple projects. Now you're forcing every project to use yaml configuration.

For now, don't load anything. Just let the `Open` function expect a config variable containing required data.
Again, if you want to try it out, don't do it in `library/main`, just create another project that will import the `postgres` pkg.

config.yml
```diff
@@ -0,0 +1,7 @@
+server:
+  host: "localhost"
+  port: 5432
+database:
+  user: "bojan"
+  pass: "bojan"
+  dbname: "library"
```
sfqi: We don't need this yet.

README.md
```diff
@@ -7,4 +7,8 @@ at the and of file, enter this line at the end of file:
 export LIBRARY='https://openlibrary.org/api/'
 Logout, and login again 
 
-We need to remove '/' at the end from https://openlibrary.org/api/ , and add it in bookPath
\ No newline at end of file
+We need to remove '/' at the end from https://openlibrary.org/api/ , and add it in bookPath
+
+For connecting to database, we must set our username and password as env var:
+export PSQLUSERNAME=*****
+export PSQLPASSWORD=*****
```
sfqi: We don't need this yet.

PR #28, opened 2019-11-05
Feature/fixing from pr24

bokimilinkovic:
I am unable to test error from db.Create, since that function always returns error as nil .. Should i just leave it for now, or change body of that function inside database.go ?

### DIFF
--- a/handler/handlers.go
+++ b/handler/handlers.go
@@ -2,7 +2,6 @@ package handler
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


sfqi:
> I am unable to test error from db.Create, since that function always returns error as nil .. Should i just leave it for now, or change body of that function inside database.go ?

Create a trello card saying: Missing test case for db.Create returning an error

handler/handlers_test.go
```diff
@@ -159,7 +150,7 @@ func TestCreate(t *testing.T) {
 		}
 	})
 	t.Run("Fetching book error", func(t *testing.T) {
-		clmock := mockClient.ClientMock{nil,
+		clmock := olmock.ClientMock{nil,
```
sfqi: `ClientMock` -> `Client`

handler/handlers.go
```diff
@@ -65,7 +66,7 @@ func (b *BookHandler) Create(w http.ResponseWriter, r *http.Request) {
 	book := b.toBook(openlibraryBook)
 
 	if err := b.Db.Create(book); err != nil {
-		http.Error(w, err.Error(), http.StatusBadRequest)
+		http.Error(w, errors.New("Error creating book in database").Error(), http.StatusBadRequest)
```
sfqi: No need for this I think.

PR #30, opened 2019-11-07
Added test scenario for book creation and invalid request body

bokimilinkovic:
https://trello.com/c/vdx4w7mn/56-add-missing-test-cases-to-create-book-handler


handler/handlers_test.go
```diff
@@ -172,6 +173,34 @@ func TestCreate(t *testing.T) {
...
+			t.Errorf("Error occured, %s", err)
+		}
+
+		rr := httptest.NewRecorder()
+
+		handler := http.HandlerFunc(bookHandler.Create)
+
+		handler.ServeHTTP(rr, req)
+		//Since we dont know what will be returned as response, I put interface
+		//We expect to be returned map[string]interface{}, in other scenario error is returned and test will fail
```
sfqi: What do you mean we don't know? We're expecting a serialized `bookResponse`.

handler/handlers.go
```diff
@@ -73,7 +73,7 @@ func (b *BookHandler) Create(w http.ResponseWriter, r *http.Request) {
 	var createBook dto.CreateBookRequest
 
 	if err := json.NewDecoder(r.Body).Decode(&createBook); err != nil {
-		errorDecodingBook(w, err)
+		errorDecodingBook(w,err)
```
sfqi: Why?

handler/handlers_test.go
```diff
@@ -127,14 +129,13 @@ func TestUpdate(t *testing.T) {
 }
 
 func TestCreate(t *testing.T) {
-	t.Run("Error decoding Book attributes", func(t *testing.T) {
-
+	t.Run("Invalid request body", func(t *testing.T) {
 		clmock := olmock.Client{nil,
 			errors.New("Error while decoding from request body"),
 		}
```
sfqi: This test case should pass without initializing the error here, since the handler should return before reaching `Fetch(isbn)`

handler/handlers_test.go
bokimilinkovic: i mean we can get response as type bookResponse, or error type ... We expect to get something different from error to pass test

handler/handlers_test.go
bokimilinkovic: you're right, it's passing without initializing error

handler/handlers_test.go
sfqi: What we're getting as the response is a bunch of bytes. We decode them into `bookResponse`. If the resulting struct is empty or there's been an error while decoding - well, that means the test should fail.

handler/handlers_test.go
```diff
@@ -2,6 +2,9 @@ package handler
 
 import (
 	"bytes"
+	"encoding/json"
+	dto2 "github.com/library/handler/dto"
```
sfqi: Guess it won't be the final version of this PR: just leave `dto`

handler/handlers_test.go
```diff
@@ -2,6 +2,9 @@ package handler
...
+	"github.com/library/openlibrary/dto"
```
sfqi: `openlibrarydto`

handler/handlers_test.go
sfqi: pretty sure you could have just gone with olmock.Client{}

handler/handlers_test.go
```diff
@@ -172,6 +174,32 @@ func TestCreate(t *testing.T) {
...
+		handler := http.HandlerFunc(bookHandler.Create)
+
+		handler.ServeHTTP(rr, req)
+
+		var response *dto2.BookResponse
+		err = json.NewDecoder(rr.Body).Decode(&response)
+		if response == nil {
+			t.Errorf("Response is nil, test is failing : %s",err)
+		}else{
+			fmt.Println("Book is successfuly created")
```
sfqi: You don't need this `else`

handler/handlers_test.go
```diff
@@ -172,6 +174,32 @@ func TestCreate(t *testing.T) {
...
+		rr := httptest.NewRecorder()
+
+		handler := http.HandlerFunc(bookHandler.Create)
+
+		handler.ServeHTTP(rr, req)
+
+		var response *dto2.BookResponse
+		err = json.NewDecoder(rr.Body).Decode(&response)
+		if response == nil {
+			t.Errorf("Response is nil, test is failing : %s",err)
```
sfqi: 1. Handler the error properly (`if err != nil...`)
2. Check the content of the response against an expected `bookResponse`. Meaning, you should initialize a `bookResponse` object with the expected field values. (tak a look at https://github.com/sfqi/library/pull/31)

handler/handlers_test.go
```diff
@@ -172,6 +174,32 @@ func TestCreate(t *testing.T) {
 			t.Errorf("Expected error to be %s, got error: %s", clmock.Err.Error(), rr.Body.String())
 		}
 	})
+	t.Run("Testing book creation", func(t *testing.T) {
+		clmock := olmock.Client{&dto.Book{},
```
sfqi: You might want to populate this `openlibrarydto.Book` object so that it actually returns some data (and then we can check that data at the end of the test).

Also, either do
```
SomeStruct{attr1, attr2}
```

or 
```
SomeStruct{
    attr1,
    attr2,
}
```

not
```
SomeStruct{attr1,
   attr2,
}
```

handler/handlers.go
sfqi: Why delete the empty space? They're nice, they make stuff easier to read.

sfqi:
You're killing me with those commit messages

handler/handlers_test.go
```diff
@@ -192,13 +193,20 @@ func TestCreate(t *testing.T) {
...
 		if response == nil {
 			t.Errorf("Response is nil, test is failing : %s",err)
-		}else{
-			fmt.Println("Book is successfuly created")
 		}
+		if response.Title!=clmock.Book.Title || response.Year !=clmock.Book.Year{
+			t.Errorf("Book attribute did not match, wanted %v , got %v",clmock.Book,response)
+		}
+
+		fmt.Println("Book is successfuly created")
```
sfqi: You also didn't need what was inside the `else`

handler/handlers_test.go
```diff
@@ -175,7 +173,10 @@ func TestCreate(t *testing.T) {
 		}
 	})
 	t.Run("Testing book creation", func(t *testing.T) {
-		clmock := olmock.Client{&dto.Book{},
+		clmock := olmock.Client{&openlibrarydto.Book{
+			Title:      "Test title",
+			Year:       "Test year",
+		},
```
sfqi: Don't openlibrary responses usually have more attributes than just those two?

handler/handlers_test.go
```diff
@@ -192,13 +193,20 @@ func TestCreate(t *testing.T) {
...
+			t.Errorf("Error decoding %s",err.Error())
+		}
 		if response == nil {
 			t.Errorf("Response is nil, test is failing : %s",err)
-		}else{
-			fmt.Println("Book is successfuly created")
 		}
+		if response.Title!=clmock.Book.Title || response.Year !=clmock.Book.Year{
+			t.Errorf("Book attribute did not match, wanted %v , got %v",clmock.Book,response)
+		}
```
sfqi: Again, take a look at https://github.com/sfqi/library/pull/31.
We create a `dto.BookResponse` var named `expectedResponse` and initialize it with the expected attributes.

handler/handlers_test.go
bokimilinkovic: it has, but I don't think its necesary to pass all of them, especialy because request structure and response structure dont have the same attributes, so we need parsing to check all of them. Title and Year are the same so i only check them to see if it's working fine

sfqi: Maybe we'd like to test the parsing as well? And the converting from model to response?

PR #33, opened 2019-11-12
Delete book endpoint added

bokimilinkovic:
https://trello.com/c/e54L4fBV/64-add-delete-book-endpoint


repository/mock/database.go
```diff
@@ -89,3 +89,14 @@ func (db *DB) Update(toUpdate *model.Book) error {
...
+func (db *DB)Delete(toDelete *model.Book)error{
+	temp := db.books
+	db.books=[]model.Book{}
+	for _, book := range temp{
+		if book.Id != toDelete.Id{
+			db.books=append(db.books,book)
+		}
+	}
+	return nil
+}
```
sfqi: Maybe you could use `findBookByID` to get the location and then do something like [this](https://stackoverflow.com/questions/37334119/how-to-delete-an-element-from-a-slice-in-golang#answer-55381756) or [this](https://stackoverflow.com/questions/37334119/how-to-delete-an-element-from-a-slice-in-golang#answer-37335777).

sfqi:
Tests?

repository/mock/database.go
sfqi: Do we need `temp` and `db.books=[]model.Book{}`?

handler/handlers.go
```diff
@@ -204,6 +204,27 @@ func (b *BookHandler) Get(w http.ResponseWriter, r *http.Request) {
...
+	}
+	book, err := b.Db.FindBookByID(id)
+	if err != nil {
+		errorFindingBook(w, err)
+		return
+	}
+	if err := b.Db.Delete(book); err != nil {
+		return
+	}
+	fmt.Fprintln(w,"Book successfuly deleted")
```
sfqi: Just return `StatusNoContent`, no need to print anything

repository/mock/database.go
sfqi: Also, let's try to be consistent with empty spaces (as in, we should use them)

handler/handlers.go
bokimilinkovic: you mean just to write fmt.Fprintln(w, http.SatusNoContent)?  I cant return anything in handler

sfqi: You figured it out in the meantime.

repository/mock/database.go
sfqi: What about these comments?

repository/mock/database.go
```diff
@@ -89,3 +89,16 @@ func (db *DB) Update(toUpdate *model.Book) error {
 	db.books[index] = *book
 	return nil
 }
+
+func (db *DB)Delete(book *model.Book)error{
+	_,loc,err:=db.findBookByID(book.Id)
+	if err != nil{
+		return err
+	}
+	db.books=remove(db.books,loc)
```
sfqi: `db.books = append(db.books[:loc], db.books[loc+1:]...)` would have worked as well, I think.

handler/handlers.go
```diff
@@ -204,6 +204,27 @@ func (b *BookHandler) Get(w http.ResponseWriter, r *http.Request) {
...
+		errorConvertingId(w, err)
+		return
+	}
+	book, err := b.Db.FindBookByID(id)
+	if err != nil {
+		errorFindingBook(w, err)
+		return
+	}
+	if err := b.Db.Delete(book); err != nil {
+		return
```
sfqi: No error handling here?

PR #34, opened 2019-11-12
Cleaned postgres pkg

bokimilinkovic:
https://trello.com/c/NePM2Rxp/59-clean-up-postgres-pkg

### DIFF
--- a/repository/postgres/postgres.go
+++ b/repository/postgres/postgres.go
@@ -17,34 +17,29 @@ type PostgresConfig struct{
 
 
 type Store struct {
-	Db db
-}
-type db struct{
-	DB *gorm.DB
+	db *gorm.DB
 }
 
 func NewStore()*Store{
-	return &Store{
-		Db:db{
-			DB:&gorm.DB{},
-		},
-	}
+	return &Store{}
 }
 
 func Open(config PostgresConfig)(*Store, error){
 	Store := NewStore()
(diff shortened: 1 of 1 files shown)


repository/postgres/postgres.go
```diff
@@ -16,25 +16,35 @@ type PostgresConfig struct{
 }
 
 
-type dbStore struct {
+type Store struct {
+	Db db
+}
+type db struct{
```
sfqi: type db *gorm.DB

repository/postgres/postgres.go
```diff
@@ -16,25 +16,35 @@ type PostgresConfig struct{
 }
 
 
-type dbStore struct {
+type Store struct {
+	Db db
```
sfqi: Also unexport this

repository/postgres/postgres.go
```diff
@@ -17,34 +17,29 @@ type PostgresConfig struct{
...
+	db *gorm.DB
 }
 
 func NewStore()*Store{
-	return &Store{
-		Db:db{
-			DB:&gorm.DB{},
-		},
-	}
+	return &Store{}
```
sfqi: How about `NewStore(db *gorm.DB) *Store`?

Then you could actually use it in `Open`.

bokimilinkovic: good idea

PR #35, opened 2019-11-12
Added created_at & updated_at fields to book table

StefanGajic:
https://trello.com/c/AyW5dUwE/63-add-createdat-updatedat-fields-to-user-table-and-model


migrations/1_create_books_table.up.sql
```diff
@@ -7,5 +7,7 @@ CREATE TABLE books (
     open_library_id         varchar(50),
     cover_id varchar(50),
     year    int ,
-    publisher varchar(255)
+    publisher varchar(255),
+    created_at time (0) not null,
+    updated_at time (0) not null
```
sfqi: I distinctly remember saying this should be done in another file.
The point of migrations is that changes can be applied incrementally. So in this case, we need an `up` file that alters the table by adding columns, and a `down` file that removes those columns.
Ideally, a timestamp would be used instead of a plain number (like 1)

Also, how about `timestampz` instead of `time`?

handler/dto/Book.go
```diff
@@ -9,12 +11,14 @@ type UpdateBookRequest struct {
...
-	Year          string `json:"year"`
+	ID            int       `json:id`
+	Title         string    `json:"title"`
+	Author        string    `json:"author"`
+	Isbn          string    `json:"isbn_10"`
+	Isbn13        string    `json:"isbn_13"`
+	OpenLibraryId string    `json:"olid"`
+	CoverId       string    `json:"cover"`
+	Year          string    `json:"year"`
+	CreatedAt     time.Time `json:created_at`
```
sfqi: Pretty sure I said not to include the timestamps in the response

repository/mock/database.go
```diff
@@ -72,6 +80,7 @@ func (db *DB) GetAllBooks() []model.Book {
 func (db *DB) Create(book *model.Book) error {
 	db.Id++
 
+	book.CreatedAt = time.Now()
```
sfqi: Set `UpdatedAt` as well

handler/dto/Book.go
```diff
@@ -11,14 +9,14 @@ type UpdateBookRequest struct {
...
+	ID            int    `json:id`
+	Title         string `json:"title"`
+	Author        string `json:"author"`
+	Isbn          string `json:"isbn_10"`
+	Isbn13        string `json:"isbn_13"`
+	OpenLibraryId string `json:"olid"`
+	CoverId       string `json:"cover"`
+	Year          string `json:"year"`
+	// CreatedAt     time.Time `json:created_at`
+	// UpdatedAt     time.Time `json:updated_at`
```
sfqi: Delete delete.

repository/mock/database.go
sfqi: Well, you could do something like
```
now := time.Now()
book.CreatedAt = now
book.UpdatedAt = now
```

I mean it's dumb, but why not.

migrations/2_alter_books_table.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE books ADD created_at timestamptz (timestamp) not null;
+ALTER TABLE books ADD updated_at timestamptz (timestamp) not null;
```
sfqi: Do we need the `(timestamp)` part? Genuinely asking.

Did you try running the migration? Up & down?

repository/postgres/postgres.go
```diff
@@ -7,7 +7,6 @@ import (
 	_ "github.com/jinzhu/gorm/dialects/postgres"
 	"github.com/sfqi/library/domain/model"
 )
-
```
sfqi: Why remove the line?

domain/model/BookModel.go
```diff
@@ -1,12 +1,16 @@
...
+	Id            int       `json:"id"`
+	Title         string    `json:"title"`
+	Author        string    `json:"author"`
+	Isbn          string    `json:"isbn_10"`
+	Isbn13        string    `json:"isbn_13"`
+	OpenLibraryId string    `json:"olid"`
+	CoverId       string    `json:"cover"`
+	Year          string    `json:"publish_date"`
+	CreatedAt     time.Time `json:created_at`
+	UpdatedAt     time.Time `json:updated_at`
```
sfqi: Can you rename this file to `model/book.go` while you're at it?

repository/mock/database.go
```diff
@@ -16,6 +17,8 @@ var books = []model.Book{
 		OpenLibraryId: "again some id",
 		CoverId:       "some cover ID",
 		Year:          "2019",
+		CreatedAt:     time.Now(),
+		UpdatedAt:     time.Now(),
```
sfqi: Here's a thought:
Maybe you could define three time vars - now, 10 secs earlier, and 15 secs earlier, and use them as timestamps.

PR #36, opened 2019-11-13
CreateBook method over postgres created

bokimilinkovic:
pq: column "isbn13" of relation "books" does not exist 
We will need to fix some issues about tagging columns in database
https://trello.com/c/Lj7GPZWu/60-implement-postgres-createbook-method

### DIFF
--- a/repository/postgres/postgres.go
+++ b/repository/postgres/postgres.go
@@ -46,3 +46,7 @@ func(store *Store)FindById(id int)(*model.Book, error){
 	return &b,nil
 }
 
+func (store *Store)CreateBook(book model.Book)(err error){
+	store.db.Save(&book) //Save will insert primary key if it doesn't exist
+	return nil
+}
--- a/repository/postgres/postgres.go
+++ b/repository/postgres/postgres.go
@@ -46,7 +46,6 @@ func(store *Store)FindById(id int)(*model.Book, error){
 	return &b,nil
 }
 
-func (store *Store)CreateBook(book model.Book)(err error){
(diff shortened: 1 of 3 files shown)


sfqi:
> pq: column "isbn13" of relation "books" does not exist
> We will need to fix some issues about tagging columns in database
> https://trello.com/c/Lj7GPZWu/60-implement-postgres-createbook-method

What's stopping you?
http://gorm.io/docs/conventions.html#Snake-Case-Column-Name

repository/postgres/postgres.go
```diff
@@ -46,3 +46,7 @@ func(store *Store)FindById(id int)(*model.Book, error){
 	return &b,nil
 }
 
+func (store *Store)CreateBook(book model.Book)(err error){
+	store.db.Save(&book) //Save will insert primary key if it doesn't exist
```
sfqi: Remove the comment.
I think you can use `Create` here.

We haven't been using named return params in this project, so let's not mix things up.

`(book *model.Book)` might be better.

And error handling is missing. In this case, adding `.Error` to the gorm command would be the way to go.

repository/postgres/postgres.go
```diff
@@ -46,3 +46,6 @@ func(store *Store)FindById(id int)(*model.Book, error){
 	return &b,nil
 }
 
+func (store *Store)CreateBook(book *model.Book)(err error){
+	return store.db.Create(&book).Error //Save will insert primary key if it doesn't exis
```
sfqi: Just delete the comment

PR #38, opened 2019-11-15
Added update book method with gorm

StefanGajic:
https://trello.com/c/SksU9UMl/61-implement-postgres-updatebook-method

### DIFF
--- a/handler/dto/book.go
+++ b/handler/dto/book.go
@@ -5,7 +5,7 @@ type CreateBookRequest struct {
 }
 type UpdateBookRequest struct {
 	Title string `json:"title"`
-	Year  string `json:"publish_date"`
+	Year  string `json:"year"`
 }
 
 type BookResponse struct {
@@ -16,5 +16,5 @@ type BookResponse struct {
 	Isbn13        string `json:"isbn_13"`
 	OpenLibraryId string `json:"olid"`
 	CoverId       string `json:"cover"`
-	Year          string `json:"publish_date"`
+	Year          string `json:"year"`
 }
--- a/handler/dto/book.go
+++ b/handler/dto/book.go
@@ -5,7 +5,7 @@ type CreateBookRequest struct {
 }
 type UpdateBookRequest struct {
 	Title string `json:"title"`
-	Year  string `json:"year"`
+	Year  string `json:"publish_date"`
 }
 
 type BookResponse struct {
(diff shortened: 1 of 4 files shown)


handler/dto/book.go
```diff
@@ -16,5 +16,5 @@ type BookResponse struct {
 	Isbn13        string `json:"isbn_13"`
 	OpenLibraryId string `json:"olid"`
 	CoverId       string `json:"cover"`
-	Year          string `json:"year"`
+	Year          string `json:"publish_date"`
```
sfqi: dtos and json tags don't have anything to with the database, no need to change these

openlibrary/dto/Book.go
```diff
@@ -5,7 +5,7 @@ type Book struct {
 	Identifier identifier `json:"identifiers"`
 	Author     []author   `json:"authors"`
 	Cover      cover      `json:"cover"`
-	Year       string     `json:"year"`
+	Year       string     `json:"publish_date"`
```
sfqi: See the other dto comment

repository/postgres/postgres.go
```diff
@@ -16,36 +16,39 @@ type PostgresConfig struct {
...
+}
+
+func (store *Store) CreateBook(book *model.Book) (err error) {
+	return store.db.Create(&book).Error
+
 }
 
-func (store *Store)CreateBook(book *model.Book)(err error){
-	return store.db.Create(&book).Error //Save will insert primary key if it doesn't exis
+func (store *Store) UpdateBook(book model.Book) (err error) {
```
sfqi: Expect a pointer to `model.Book` as the argument. Don't use a named return parameter.

repository/postgres/postgres.go
```diff
@@ -16,36 +16,39 @@ type PostgresConfig struct {
...
-	if err := store.db.First(&b,id).Error;err!=nil{
-		return nil,err
+	if err := store.db.First(&b, id).Error; err != nil {
+		return nil, err
 	}
-	return &b,nil
+	return &b, nil
+}
+
+func (store *Store) CreateBook(book *model.Book) (err error) {
```
sfqi: Pretty sure I said not to use named return parameters @bokimilinkovic

repository/postgres/postgres.go
sfqi: @bokimilinkovic @StefanGajic To be clear, a named return parameter is per example the `(err error)` part of the function signature. `err` is the name and `error` is the type of the return parameter. What I'd like is to use the same style of defining return parameters as the one we've been using so far.

@StefanGajic Any objections to having a `book *model.Book` instead of `book model.Book` as the argument?

PR #39, opened 2019-11-15
Deleted json tags, added gorm tags

bokimilinkovic:
https://trello.com/c/4J7eoAHO/66-remove-json-tags-from-book-model

### DIFF
--- a/domain/model/book.go
+++ b/domain/model/book.go
@@ -10,7 +10,7 @@ type Book struct {
 	Isbn13        string
 	OpenLibraryId string
 	CoverId       string
-	Year          string    `gorm:"-"`
+	Year          string    `gorm:"year"`
 	CreatedAt     time.Time
 	UpdatedAt     time.Time
 }
--- a/domain/model/book.go
+++ b/domain/model/book.go
@@ -6,11 +6,11 @@ type Book struct {
 	Id            int
 	Title         string
 	Author        string
-	Isbn          string    `gorm:"colum:isbn_10"`
-	Isbn13        string    `gorm:"column:isbn_13"`
-	OpenLibraryId string    `gorm:"column:open_library_id"`
-	CoverId       string    `gorm:"column:cover_id"`
-	Year          string    `gorm:"column:publish_date"`
-	CreatedAt     time.Time `gorm:"column:created_at"`
-	UpdatedAt     time.Time `gorm:"column:updated_at"`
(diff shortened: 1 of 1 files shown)


domain/model/book.go
```diff
@@ -3,14 +3,14 @@ package model
...
-	Year          string    `json:"publish_date"`
-	CreatedAt     time.Time `json:created_at`
-	UpdatedAt     time.Time `json:updated_at`
+	Id            int
+	Title         string
+	Author        string
+	Isbn          string    `gorm:"colum:isbn_10"`
+	Isbn13        string    `gorm:"column:isbn_13"`
+	OpenLibraryId string    `gorm:"column:open_library_id"`
+	CoverId       string    `gorm:"column:cover_id"`
```
sfqi: I don't think we need every one of those tags. See http://gorm.io/docs/conventions.html#Snake-Case-Column-Name

domain/model/book.go
```diff
@@ -3,14 +3,14 @@ package model
...
-	CreatedAt     time.Time `json:created_at`
-	UpdatedAt     time.Time `json:updated_at`
+	Id            int
+	Title         string
+	Author        string
+	Isbn          string
+	Isbn13        string
+	OpenLibraryId string
+	CoverId       string
+	Year          string    `gorm:"-"`
```
sfqi: Don't we have a `year` column?

bokimilinkovic: I didn't get what we were doing with publish_date in domain model previously .. but I see now stefan changed that to 'year' , so okay i will change it.

bokimilinkovic: but year is int in db scheme as i recall, how can this be string then?

sfqi: You can create a trello card for that (e.g. "modify `model.Book.Year` type to int")

domain/model/book.go
```diff
@@ -3,14 +3,14 @@ package model
...
-	CreatedAt     time.Time `json:created_at`
-	UpdatedAt     time.Time `json:updated_at`
+	Id            int
+	Title         string
+	Author        string
+	Isbn          string
+	Isbn13        string
+	OpenLibraryId string
+	CoverId       string
+	Year          string    `gorm:"year"`
```
sfqi: I don't think this is needed. If the attribute is called `Yead`, gorm automatically maps it to column `year`

PR #40, opened 2019-11-18
Adding middleware that logs the request body

StefanGajic:
https://trello.com/c/NhVBjutN/69-add-middleware-that-logs-the-request-body

### DIFF
--- a/middleware/requestdump.go
+++ b/middleware/requestdump.go
@@ -1,0 +1,20 @@
+package middleware
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httputil"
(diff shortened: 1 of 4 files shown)


middleware/requestdump.go
```diff
@@ -0,0 +1,20 @@
+package middleware
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httputil"
+)
+
+func LoggingMiddleware(next http.Handler) http.Handler {
```
sfqi: How about `BodyDump`? The package is already called `middleware`.

PR #42, opened 2019-11-20
Made logrus object and wrote the log nicely

StefanGajic:
https://trello.com/c/2c3fECVA/70-initialize-a-logrus-object-and-use-it-to-log-from-bodydump-middleware

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/sfqi/library/openlibrary"
+	"github.com/sirupsen/logrus"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -22,13 +23,20 @@ func main() {
 		Olc: olc,
 	}
 
+	log := logrus.New()
+	log.SetFormatter(&logrus.TextFormatter{
+		FullTimestamp: true,
+	})
```
sfqi: Let's move all this to a separate package (`log`). We'll probably end up adding more stuff to it, so it will be messy if it's all here.

PR #43, opened 2019-11-22
Added Logger interface

StefanGajic:
https://trello.com/c/k1y7gcCc/74-define-a-logger-interface-in-the-log-package

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -13,5 +13,6 @@ func New() *logrus.Logger {
 }
 
 type Logger interface {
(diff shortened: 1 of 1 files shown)


log/log.go
```diff
@@ -11,3 +11,7 @@ func New() *logrus.Logger {
 	})
 	return logger
 }
+
+type Logger interface {
+	New() *logrus.Logger
+}
```
bokimilinkovic: Such a good job!

PR #44, opened 2019-11-22
Added interface, but error is found

bokimilinkovic:
I cannot find what is the error in main.go . I created DataBase interface, and declared methods we are going to use. 
BookHandler will no more have 'hardcoded' database, but DataBase which i created.
Store from postgres.go inherit all methods from DataBase interface, as so DB from inmemory.go
But when in main.go , I create handler , and set instance  DB (from inmemory.go) as a value of DataBase, it says: "Cannot set  db(type *DB) as type DataBase". But when I do the same thing with store, it is working. Why is it not working with DB from inmemory.go?


main.go
```diff
@@ -10,17 +11,31 @@ import (
...
+	if err != nil{
+		panic(err)
+	}
 
-	db := inmemory.NewDB()
-	bookHandler := handler.BookHandler{
-		Db:  db,
+	bookHandler := &handler.BookHandler{
+		DataBase: store,
 		Olc: olc,
```
bokimilinkovic: Here is the problem, with 'store' it is working, but when I create
       db := inmemory.NewDb( ) , and set:
	bookHandler := &handler.BookHandler{
		DataBase: db
I got error saying "cannot use db(type *DB) as DataBase

handler/handlers.go
```diff
@@ -14,25 +14,28 @@ import (
 
 	openlibrarydto "github.com/sfqi/library/openlibrary/dto"
 
-	"github.com/sfqi/library/repository/inmemory"
+
 
 )
 
+type DataBase interface{
```
sfqi: I would call this `store` (no exporting)

handler/handlers.go
```diff
@@ -14,25 +14,28 @@ import (
...
+	CreateBook(*model.Book) error
+	UpdateBook(*model.Book) error
+	FindAllBooks() ([]*model.Book, error)
+	GetLastId()int
+	DeleteBook(*model.Book)error
+}
+
 type BookHandler struct {
-	Db  *inmemory.DB
+	DataBase
```
sfqi: Gotta name it - `db database` or maybe `store store`

handler/handlers.go
```diff
@@ -129,7 +134,7 @@ func (b *BookHandler) toBook(book *openlibrarydto.Book) (bm *model.Book) {
 	}
 
 	bookToAdd := model.Book{
-		Id:            b.Db.Id,
+		Id:            b.GetLastId(),
```
sfqi: We shouldn't need this, let the `inmemory` database set the id (like gorm does)

repository/inmemory/database.go
```diff
@@ -12,6 +12,15 @@ var timeNow = time.Now()
...
 
+type DataBase interface{
+	FindById(int) (*model.Book, error)
+	CreateBook(*model.Book) error
+	UpdateBook(*model.Book) error
+	FindAllBooks() ([]*model.Book, error)
+	GetLastId()int
+	DeleteBook(*model.Book)error
+}
+
```
sfqi: I don't think you need this here, it's already defined in `handler`

repository/postgres/postgres.go
```diff
@@ -16,12 +16,22 @@ type PostgresConfig struct {
...
 }
 
+type DataBase interface{
+	FindById(int) (*model.Book, error)
+	CreateBook(*model.Book) error
+	UpdateBook(*model.Book) error
+	FindAllBooks() ([]*model.Book, error)
+	GetLastId()int
+	DeleteBook(*model.Book)error
+}
```
sfqi: Same comment as for 'inmemory'

repository/postgres/postgres.go
```diff
@@ -60,3 +70,17 @@ func (store *Store) FindAllBooks() ([]*model.Book, error) {
 	}
 	return books, nil
 }
+
+func(store *Store)GetLastId()int{
+	var book model.Book
+	store.db.Last(&model.Book{}).Scan(&book)
+	return book.Id
+}
```
sfqi: We should probably try to avoid this

main.go
sfqi: It probably doesn't implement something required by the interface

sfqi:
Just run go fmt

PR #46, opened 2019-11-29
Initial repository mock

bokimilinkovic:


sfqi:
Well I meant also use the mock in the tests :)

handler/handlers_test.go
```diff
@@ -2,22 +2,67 @@ package handler
...
+
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
 )
 
-var db = inmemory.NewDB()
+var db = &mock.Store{
```
sfqi: I think it's better if the mock fields are initialized in every test (otherwise you might end up with tests that depend on each other, which tends to be annoying). You can maybe define the book objects as global vars that don't get changed, and then use them initialize the mock db when needed.

handler/handlers_test.go
```diff
@@ -2,22 +2,67 @@ package handler
...
+			Id:            1,
+			Title:         "some title",
+			Author:        "some author",
+			Isbn:          "some isbn",
+			Isbn13:        "some isbon13",
+			OpenLibraryId: "again some id",
+			CoverId:       "some cover ID",
+			Year:          2019,
+			CreatedAt:     time.Time{},
+			UpdatedAt:     time.Time{},
```
sfqi: Is this the exact same object as the previous one? Surely there's a way to refactor that.

handler/handlers_test.go
```diff
@@ -2,22 +2,60 @@ package handler
...
+			Isbn13:        "other isbon13",
+			OpenLibraryId: "other some id",
+			CoverId:       "other cover ID",
+			Year:          2019,
+			CreatedAt:     time.Time{},
+			UpdatedAt:     time.Time{},
+		},
+	}
+	 return book1,books
+}
```
sfqi: Pretty sure you could just return the slice since it contains the book (book1).

So the fun could just be called `books`. But what you ended up doing is defining a global var :)

handler/handlers_test.go
```diff
@@ -2,22 +2,60 @@ package handler
...
+
+var book, books = initializeAttributes()
+var db = mock.NewStore(book,books,nil)
```
sfqi: Do you need these?

handler/handlers_test.go
```diff
@@ -160,15 +187,16 @@ func TestUpdate(t *testing.T) {
 		if err != nil {
 			t.Errorf("Error occured, %s", err)
 		}
-
+		db.Err=errors.New("Book with given Id can not be found")
+		//db.Err = errors.New("Book with given Id can not be found")
```
sfqi: ?

handler/handlers_test.go
```diff
@@ -2,22 +2,60 @@ package handler
...
+			UpdatedAt:     time.Time{},
+		},
+	}
+	 return book1,books
+}
+
+var book, books = initializeAttributes()
+var db = mock.NewStore(book,books,nil)
+
 var bookHandler BookHandler = BookHandler{
```
sfqi: This shouldn't be a global var, it should be defined in every test.

PR #47, opened 2019-12-03
Feature/creating user model

bokimilinkovic:
https://trello.com/c/d3DbwkXr/77-create-user-model-inmemory-data

### DIFF
--- a/migrations/3_create_user_table.up.sql
+++ b/migrations/3_create_user_table.up.sql
@@ -1,0 +1,9 @@
+CREATE TABLE users(
+    id serial primary key,
+    email varchar(255) not null unique,
+    name varchar(20),
+    lastname varchar(50),
+    created_at timestamptz not null,
+    updated_at timestamptz
+);
+
--- a/repository/inmemory/userDB.go
+++ b/repository/inmemory/userDB.go
@@ -1,0 +1,96 @@
+package inmemory
+
+import (
+	"fmt"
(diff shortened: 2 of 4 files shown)


repository/inmemory/userDB.go
```diff
@@ -0,0 +1,96 @@
...
+		Id:        2,
+		Email:     "drugi@email.com",
+		Name:      "ime2",
+		LastName:  "prezime2",
+		CreatedAt: time.Now(),
+		UpdatedAt: time.Now(),
+	},
+}
+
+type UserDb struct{
```
sfqi: We should probably use the same db for every table.
Table-specific structs can be embedded in the db struct later on, to provide namespacing, but if we already have a `Db` (which should be `DB`) for books, we can't have a `UserDb` for users, it's not consistent.

migrations/3_create_user_table.up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE users(
+    id serial primary key,
+    email varchar(255) not null unique,
+    name varchar(20),
+    lastname varchar(50),
```
sfqi: Just use 255, for both.

sfqi:
G
O

F
M
T

migrations/3_create_user_table.up.sql
sfqi: ahem

PR #51, opened 2019-12-05
Godotenv implemented

bokimilinkovic:
https://trello.com/c/EmCnzEgM/84-pass-configuration-params-to-app-through-dotenv-file


bokimilinkovic:
I know this .env file should go in gitignore, just for the first time to be easier for Stefan to get that file and change to his credentials.

main.go
```diff
@@ -2,30 +2,37 @@ package main
...
+	fmt.Println("Successfully read .env file")
+}
 
 func main() {
 	openLibraryUrl := os.Getenv("LIBRARY")
 	olc := openlibrary.NewClient(openLibraryUrl)
 	config := postgres.PostgresConfig{
-		Host:     "localhost",
+		Host:     os.Getenv("DB_HOST"),
 		Port:     5432,
```
sfqi: Port as well

StefanGajic:
> I know this .env file should go in gitignore, just for the first time to be easier for Stefan to get that file and change to his credentials.

Thank you so much I really appreciate it, you are so kind and selfless.
It would be great if you also add me as your reviewer so I can at least get updates of your commits and changes, but I don't know if I am asking too much?

main.go
```diff
@@ -2,30 +2,37 @@ package main
...
 	"github.com/sfqi/library/middleware"
 )
 
-
+func init(){
+	e := godotenv.Load()
+	if e != nil{
+		panic(e)
+	}
+	fmt.Println("Successfully read .env file")
```
sfqi: We don't need to print this.

Do we need the `init` if we're using autoload?

In any case, with or without autoload, I'd prefer it if we loaded the vars inside `main()`

Also, it's always 'err', not 'e'

.env
```diff
@@ -0,0 +1,5 @@
+DB_USER=bojan
+DB_PASSWORD=bojan
+DB_HOST=localhost
+DB_DIALECT=postgres
+DB_DATABASE=library
```
sfqi: The convention is to have an .env.dist file that looks like this, just with example values instead of 'bojan'.
Then everyone creates the .env file locally, based on the dist one

main.go
```diff
@@ -2,30 +2,37 @@ package main
...
+}
 
 func main() {
 	openLibraryUrl := os.Getenv("LIBRARY")
```
sfqi: This should go in the env file as well

.env
```diff
@@ -0,0 +1,7 @@
...
+DB_PORT=5432
+LIBRARY_URL=https://openlibrary.org/api
```
sfqi: Since you created the .env.dist file, we don't need this one.

.env.dist
```diff
@@ -0,0 +1,7 @@
+DB_USER=user
+DB_PASSWORD=pass
+DB_HOST=localhost
+DB_DIALECT=postgres
+DB_DATABASE=dbname
+DB_PORT=port
+LIBRARY_URL=https://openlibrary.org/api
```
sfqi: Make it `OPEN_LIBRARY_URL`

main.go
```diff
@@ -16,16 +18,22 @@ import (
...
-	openLibraryUrl := os.Getenv("LIBRARY")
+	err := godotenv.Load()
+	if err != nil{
+		panic(err)
+	}
+
+	openLibraryUrl := os.Getenv("LIBRARY_URL")
+	fmt.Println(openLibraryUrl)
 	olc := openlibrary.NewClient(openLibraryUrl)
+	port,err := strconv.Atoi(os.Getenv("DB_PORT"))
```
sfqi: Error handling?

main.go
```diff
@@ -16,16 +18,22 @@ import (
...
 func main() {
-	openLibraryUrl := os.Getenv("LIBRARY")
+	err := godotenv.Load()
+	if err != nil{
+		panic(err)
+	}
+
+	openLibraryUrl := os.Getenv("LIBRARY_URL")
+	fmt.Println(openLibraryUrl)
 	olc := openlibrary.NewClient(openLibraryUrl)
```
sfqi: I mean, these 3 lines could be just one
`olc := openlibrary.NewClient(os.Getenv("LIBRARY_URL"))`

.env.dist
```diff
@@ -0,0 +1,7 @@
+DB_USER=user
+DB_PASSWORD=pass
+DB_HOST=localhost
+DB_DIALECT=postgres
+DB_DATABASE=dbname
```
sfqi: `DB_NAME` kinda makes more sense.

.env.dist
```diff
@@ -0,0 +1,7 @@
+DB_USER=user
+DB_PASSWORD=pass
+DB_HOST=localhost
+DB_DIALECT=postgres
+DB_NAME=dbname
+DB_PORT=port
```
sfqi: This, I'm guessing, should be a number

.gitignore
```diff
@@ -2,5 +2,7 @@
 .vscode/*
 .vscode/launch.json
 exp
+.env
+./.env
```
sfqi: Do we need both?

bokimilinkovic: precautious :smile:

PR #52, opened 2019-12-06
Moved every test to separate packages

StefanGajic:
https://trello.com/c/DCOMa1Qj/80-make-sure-that-all-tests-are-in-separate-pkgs


client_test/client_test.go
```diff
@@ -1,6 +1,7 @@
-package openlibrary
+package client_test
```
sfqi: openlibrary_test

handler_test/handlers_test.go
```diff
@@ -54,7 +55,7 @@ func initializeBooks() []*model.Book {
 	return books
 }
 
-var bookHandler BookHandler = BookHandler{
+var bookHandler handler.BookHandler = handler.BookHandler{
```
sfqi: I don't think we need the first `handler.BookHandler`

openlibrary/client.go
```diff
@@ -17,19 +17,19 @@ const formatParams = "&format=json&jscmd=data"
 var fetchBookPath = bookPath + bibkeys + formatParams
 
 type Client struct {
-	url string
+	Url string
```
sfqi: Instead of exporting the `url` field, we could always use the `NewClient` constructor when initializing a new client.

PR #56, opened 2019-12-10
Parse publish_date to int and getting only year

StefanGajic:
https://trello.com/c/liM0EO2Y/68-properly-parse-openlibrary-publishing-date

### DIFF
--- a/handler/handlers.go
+++ b/handler/handlers.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/sfqi/library/domain/model"
@@ -125,6 +127,15 @@ func (b *BookHandler) toBook(book *openlibrarydto.Book) (bm *model.Book) {
 	if book.Author != nil {
 		author = book.Author[0].Name
 	}
+	yearRgx := regexp.MustCompile(`[0-9]{4}`)
+
+	yearRgx.MatchString(book.Year)
+	yearString := yearRgx.FindString(book.Year)
+
+	year, err := strconv.Atoi(yearString)
+	if err != nil {
+		fmt.Println("error while converting year from string to int", err)
+	}
 
 	bookToAdd := model.Book{
 		Title:         book.Title,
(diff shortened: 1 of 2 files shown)


handler/handlers.go
```diff
@@ -125,6 +127,15 @@ func (b *BookHandler) toBook(book *openlibrarydto.Book) (bm *model.Book) {
 	if book.Author != nil {
 		author = book.Author[0].Name
 	}
+	yearRgx := regexp.MustCompile(`[0-9]{4}`)
```
sfqi: `MustCompile` panic if there's an error. And it's not necessary to compile every time the handler is invoked. Probably better to run it once while compiling - so not in a function, but as a package-level var.

handler/handlers.go
```diff
@@ -125,6 +127,15 @@ func (b *BookHandler) toBook(book *openlibrarydto.Book) (bm *model.Book) {
...
+
+	yearRgx.MatchString(book.Year)
+	yearString := yearRgx.FindString(book.Year)
```
sfqi: Do we need both the `MatchString` and the `FindString` command?

handler/handlers.go
```diff
@@ -125,6 +127,15 @@ func (b *BookHandler) toBook(book *openlibrarydto.Book) (bm *model.Book) {
...
+
+	year, err := strconv.Atoi(yearString)
+	if err != nil {
+		fmt.Println("error while converting year from string to int", err)
```
sfqi: You want to return the error, not just print it.

handler/handlers.go
```diff
@@ -127,14 +129,13 @@ func (b *BookHandler) toBook(book *openlibrarydto.Book) (bm *model.Book) {
...
 	}
-	yearRgx := regexp.MustCompile(`[0-9]{4}`)
 
-	yearRgx.MatchString(book.Year)
 	yearString := yearRgx.FindString(book.Year)
 
 	year, err := strconv.Atoi(yearString)
 	if err != nil {
 		fmt.Println("error while converting year from string to int", err)
+		return
```
sfqi: return **the error**

PR #57, opened 2019-12-11
testify.Mock in handlers_test

StefanGajic:
https://trello.com/c/Bkk8s2Pm/89-use-testifymock

### DIFF
--- a/handler_test/handlers_test.go
+++ b/handler_test/handlers_test.go
@@ -8,15 +8,12 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/sfqi/library/handler"
 
-	"github.com/stretchr/testify/assert"
-
-	"encoding/json"
+	"github.com/stretchr/testify/mock"
 
 	"github.com/sfqi/library/domain/model"
-	"github.com/sfqi/library/handler/dto"
-	openlibrarydto "github.com/sfqi/library/openlibrary/dto"
 	olmock "github.com/sfqi/library/openlibrary/mock"
-	"github.com/sfqi/library/repository/mock"
+
+	rmock "github.com/sfqi/library/repository/mock"
 
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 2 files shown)


handler_test/handlers_test.go
```diff
@@ -25,8 +21,71 @@ import (
...
+}
+
+func (sm *StoreMock) FindAllBooks() ([]*model.Book, error) {
+	args := sm.Called()
+
+	err := args.Error(1)
+	if err != nil {
+		return nil, err
+	}
+	book := args.Get(0).([]*model.Book)
```
bokimilinkovic: 'books' since its a slice

handler_test/handlers_test.go
```diff
@@ -121,56 +126,42 @@ func TestUpdate(t *testing.T) {
...
-			t.Errorf("Status code differs. Expected %d. Got %d", http.StatusOK, status)
+		mockClient := StoreMock{}
+		book := model.Book{
+			Title:         "test title",
+			Author:        "other author",
+			Isbn:          "other isbn",
+			Isbn13:        "other isbon13",
+			OpenLibraryId: "other some id",
+			CoverId:       "other cover ID",
+			Year:          2019,
```
bokimilinkovic: this should be the same structure as UpdateBookRequest from dto

handler_test/handlers_test.go
```diff
@@ -234,138 +225,92 @@ func TestCreate(t *testing.T) {
...
-		rr := httptest.NewRecorder()
-
-		handler := http.HandlerFunc(bookHandler.Create)
-
-		handler.ServeHTTP(rr, req)
-
-		bookExpected := dto.BookResponse{
-			0,
+		mockClient := StoreMock{}
+		book := model.Book{
```
bokimilinkovic: this should be openlibrarydto.Book, since we want to send the same structure to be parsed, as we get from real request

handler_test/handlers_test.go
```diff
@@ -234,138 +225,92 @@ func TestCreate(t *testing.T) {
...
+			"War and Peace (Penguin Classics)",
+			"Tolstoy",
+			"0140447938",
+			"9780140447934",
+			"OL7355422M",
+			"5049015",
+			2019,
+			"Penguin Books",
+			time.Time{},
+			time.Time{},
```
bokimilinkovic: delete this

handler_test/handlers_test.go
```diff
@@ -234,138 +225,92 @@ func TestCreate(t *testing.T) {
...
-			OpenLibraryId: "again some id",
-			CoverId:       "some cover ID",
-			Year:          2019,
-		}
-		var response dto.BookResponse
-		err = json.NewDecoder(rr.Body).Decode(&response)
+	t.Run("No such book ID in the database", func(t *testing.T) {
+		mockClient := StoreMock{}
+		mockClient.On("FindBookById", 5).Return(nil, errors.New("Can't find the book with that ID"))
+		mockClient.FindBookById(5)
```
bokimilinkovic: what if we pass argument that is not integer?

handler_test/handlers_test.go
```diff
@@ -234,138 +225,92 @@ func TestCreate(t *testing.T) {
...
+			Author:        "other author",
+			Isbn:          "other isbn",
+			Isbn13:        "other isbon13",
+			OpenLibraryId: "other some id",
+			CoverId:       "other cover ID",
+			Year:          2019,
+		}
 
-		handler.ServeHTTP(rr, req)
+		mockClient.On("DeleteBook", &book).Return(model.Book{
```
bokimilinkovic: We don't want model.Book to be returned after successfully deleting.

PR #58, opened 2019-12-11
Mocking client using testify.Mock

bokimilinkovic:
https://trello.com/c/Bkk8s2Pm/89-use-testifymock

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,7 +7,6 @@ import (
 	"net/http"
 	"os"
 	"strconv"
-
 	_ "github.com/joho/godotenv/autoload"
 	"github.com/sfqi/library/log"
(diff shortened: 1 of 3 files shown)


sfqi:
@bokimilinkovic @StefanGajic Wait, if I'm getting this right, we're testing the openlibrary client by mocking the openlibrary client?

StefanGajic:
> @bokimilinkovic @StefanGajic Wait, if I'm getting this right, we're testing the openlibrary client by mocking the openlibrary client?

We should do something like [this](https://tutorialedge.net/golang/improving-your-tests-with-testify-go/#a-mocking-example), right?

PR #59, opened 2019-12-11
Add pre-commit githook

sfqi:
It runs go fmt on every modified file before committing.


sfqi:
Make sure to run the commands from the readme after pulling the changes.

PR #61, opened 2019-12-13
.env added in gitigore

bokimilinkovic:


.gitignore
```diff
@@ -2,7 +2,7 @@
 .vscode/*
 .vscode/launch.json
 exp
-.env
+/*.env
```
sfqi: try `/.env` instead

PR #62, opened 2019-12-13
Feature/testify.mock

bokimilinkovic:


sfqi:
1. Missing database error scenarios (`On("MethodName", arg).Return(nil, databaseError)`)
2. In some places, missing error handling when json decoding responses
3. Make sure there are no missing tests covering errors while retrieving book from context

handler_test/handlers_test.go
```diff
@@ -74,24 +63,19 @@ func TestIndex(t *testing.T) {
...
 			Title:         "other title",
 			Author:        "other author",
 			Isbn:          "other isbn",
 			Isbn13:        "other isbon13",
 			OpenLibraryId: "other some id",
 			CoverId:       "other cover ID",
 			Year:          2019,
-		},
-	}
+		}}, errors.New("Error finding all books"))
```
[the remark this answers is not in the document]
bokimilinkovic: I was about to ask you for 'retrieving book from context' test... I was not sure if I need to also check that in handlers, since we are setting context in bookloader middleware, so I assumed that I need to check in bookloader_tests if book successfully added to context.

handler_test/handlers_test.go
```diff
@@ -125,7 +127,7 @@ func TestUpdate(t *testing.T) {
 			OpenLibraryId: "other some id",
 			CoverId:       "other cover ID",
 			Year:          2019,
-		}).Return(nil)
+		}).Return(nil, errors.New("Error updating book"))
```
bokimilinkovic: UpdateBook only returns error, if there is one

handler_test/handlers_test.go
```diff
@@ -271,7 +296,7 @@ func TestCreate(t *testing.T) {
 			OpenLibraryId: "OL7355422M",
 			CoverId:       "5049015",
 			Year:          2007,
-		}).Return(nil)
+		}).Return(nil, errors.New("Error creating book"))
```
bokimilinkovic: Create also returns only error

handler_test/handlers_test.go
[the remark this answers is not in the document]
sfqi: @bokimilinkovic Well we are talking about **unit** tests. Meaning, the handler unit tests don't know anything about the middleware - what we are testing is just the handler behavior.
But also consider this: testing the middleware means that we're kinda sure that **setting** the book in the context if working ok. **Retrieving** the book from the context is a completely different operation. What if there's a coding mistake in the type assertion done after reading from context? Or if the error isn't properly handled? Middleware tests won't catch that.

handler_test/handlers_test.go
sfqi: Then I'm assuming these tests will panic when run.

handler_test/handlers_test.go
```diff
@@ -260,6 +353,52 @@ func TestCreate(t *testing.T) {
...
+				ISBN10:      []string{"0140447938"},
+				ISBN13:      []string{"9780140447934"},
+				Openlibrary: []string{"OL7355422M"},
+			},
+			Author: []openlibrarydto.Author{
+				{Name: "Tolstoy"},
+			},
+			Cover: openlibrarydto.Cover{"https://covers.openlibrary.org/b/id/5049015-S.jpg"},
+			Year:  "2007",
+		},
```
sfqi: Can we trim this so that only the title remains? Since we don't really care about the openlibrary response in this scenario (it's tested in the one without errors)

bokimilinkovic: yes, now I see we missed to protect our self from year to be empty string, so it fails to covert it to int.

sfqi: okay, leave it as is for now, I'll make a trello card for this issue

PR #65, opened 2019-12-24
Changed name from handlers to handler and moved tests to smae package…

StefanGajic:
https://trello.com/c/q092eEJG/91-fix-names-and-test-pkgs


sfqi:
Are the handler tests passing?

PR #68, opened 2019-12-24
Create method and tests

bokimilinkovic:
https://trello.com/c/RIHzYblx/98-add-book-interactor-create-method-with-tests

### DIFF
--- a/interactor/book.go
+++ b/interactor/book.go
@@ -1,8 +1,13 @@
 package interactor
 
 import (
+	"fmt"
 	"github.com/sfqi/library/domain/model"
+	"github.com/sfqi/library/handler/dto"
 	openlibrarydto "github.com/sfqi/library/openlibrary/dto"
+	"regexp"
+	"strconv"
+	"strings"
 )
 
 type store interface {
@@ -22,6 +27,8 @@ type Book struct {
 	openlib openlibraryClient
 }
 
+var yearRgx = regexp.MustCompile(`[0-9]{4}`)
+
 func NewBook(store store, olc openlibraryClient) *Book {
 	return &Book{
 		store:   store,
@@ -34,9 +41,20 @@ func (b *Book) FindAll() ([]*model.Book, error) {
 	return nil, nil
 }
 
-func (b *Book) Create(*model.Book) error {
+func (b *Book) Create(bookRequest dto.CreateBookRequest) (*model.Book, error) {
+	openLibraryBook, err := b.openlib.FetchBook(bookRequest.ISBN)
+	if err != nil {
+		fmt.Println("error while fetching book: ", err)
+		return nil, err
+	}
 
-	return nil
+	book := b.toBook(openLibraryBook)
+
+	if err := b.store.CreateBook(book); err != nil {
+		fmt.Println("Error creating book in database : " + err.Error())
+		return nil, err
+	}
+	return book, nil
 }
 
 func (b *Book) Update(*model.Book) error {
@@ -52,4 +70,52 @@ func (b *Book) FindById(int) (*model.Book, error) {
 func (b *Book) Delete(*model.Book) error {
 
 	return nil
+}
+
+func (b *Book) toBook(book *openlibrarydto.Book) (bm *model.Book) {
+	isbn10 := ""
+	if book.Identifier.ISBN10 != nil {
+		isbn10 = book.Identifier.ISBN10[0]
+	}
+	isbn13 := ""
+	if book.Identifier.ISBN13 != nil {
+		isbn13 = book.Identifier.ISBN13[0]
+	}
+
+	CoverId := ""
+	if book.Cover.Url != "" {
+		part1 := strings.Split(book.Cover.Url, "/")[5]
+		part2 := strings.Split(part1, ".")[0]
+		CoverId = strings.Split(part2, "-")[0]
+	}
+	libraryId := ""
+	if book.Identifier.Openlibrary != nil {
+		libraryId = book.Identifier.Openlibrary[0]
+	}
+	author := ""
+	if book.Author != nil {
+		author = book.Author[0].Name
+	}
+
+	year := 0
+	var err error
+	yearString := yearRgx.FindString(book.Year)
+	if yearString != "" {
+		year, err = strconv.Atoi(yearString)
+		if err != nil {
+			fmt.Println("error while converting year from string to int", err)
+			return nil
+		}
+	}
+
+	bookToAdd := model.Book{
+		Title:         book.Title,
+		Author:        author,
+		Isbn:          isbn10,
+		Isbn13:        isbn13,
+		OpenLibraryId: libraryId,
+		CoverId:       CoverId,
+		Year:          year,
+	}
+	return &bookToAdd
 }
--- a/interactor/book.go
+++ b/interactor/book.go
@@ -47,7 +47,10 @@ func (b *Book) Create(bookRequest dto.CreateBookRequest) (*model.Book, error) {
 		return nil, err
 	}
 
-	book := toBook(openLibraryBook)
+	book, err := toBook(openLibraryBook)
+	if err != nil {
+		return nil, err
+	}
 
 	if err := b.store.CreateBook(book); err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


interactor/book_test.go
```diff
@@ -1,8 +1,54 @@
...
+	t.Run("Error while fetching book", func(t *testing.T) {
+		var db = &mock.Store{}
+		clmock := &olmock.Client{}
+
+		clmock.On("FetchBook", "0140447938222").Return(nil, errors.New("Error while fetching book: "))
+		book := interactor.NewBook(db, clmock)
+		bookRequest := dto.CreateBookRequest{ISBN: "0140447938222"}
+		bookResponse, err := book.Create(bookRequest)
+		assert.Nil(bookResponse)
+		require.Contains(err.Error(), "Error while fetching book: ")
```
bokimilinkovic: Do i need both?

sfqi: yes

interactor/book_test.go
```diff
@@ -1,8 +1,54 @@
...
 
 func TestBook_Create(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+	t.Run("Error while fetching book", func(t *testing.T) {
+		var db = &mock.Store{}
+		clmock := &olmock.Client{}
+
+		clmock.On("FetchBook", "0140447938222").Return(nil, errors.New("Error while fetching book: "))
+		book := interactor.NewBook(db, clmock)
```
sfqi: `bookInteractor :=`

interactor/book_test.go
```diff
@@ -1,8 +1,54 @@
...
+	})
+	t.Run("Book successfully saved in database", func(t *testing.T) {
+		var db = &mock.Store{}
+		clmock := &olmock.Client{}
+		clmock.On("FetchBook", "0140447938").Return(&openlibrarydto.Book{Title: "War and Peace (Penguin Classics)"}, nil)
+		db.On("CreateBook", &model.Book{Title: "War and Peace (Penguin Classics)"}).Return(nil)
+		book := interactor.NewBook(db, clmock)
+		bookRequest := dto.CreateBookRequest{ISBN: "0140447938"}
+		bookResponse, err := book.Create(bookRequest)
+		assert.NotNil(bookResponse)
```
sfqi: Well you want to check the exact content of the returned object.

Also, see comments above.

interactor/book_test.go
```diff
@@ -1,8 +1,54 @@
...
+		book := interactor.NewBook(db, clmock)
+		bookRequest := dto.CreateBookRequest{ISBN: "0140447938"}
+		bookResponse, err := book.Create(bookRequest)
+		assert.Nil(bookResponse)
+		require.Contains(err.Error(), "Error saving book in database")
+	})
+	t.Run("Book successfully saved in database", func(t *testing.T) {
+		var db = &mock.Store{}
+		clmock := &olmock.Client{}
+		clmock.On("FetchBook", "0140447938").Return(&openlibrarydto.Book{Title: "War and Peace (Penguin Classics)"}, nil)
```
sfqi: Let's return a full `openlibrarydto.Book` object, so that toBook is tested as well (the way it was tested in the handler)

interactor/book_test.go
sfqi: or `b := `

interactor/book.go
```diff
@@ -53,3 +69,51 @@ func (b *Book) Delete(*model.Book) error {
...
+	yearString := yearRgx.FindString(book.Year)
+	if yearString != "" {
+		year, err = strconv.Atoi(yearString)
+		if err != nil {
+			fmt.Println("error while converting year from string to int", err)
+			return nil
+		}
+	}
+
+	bookToAdd := model.Book{
```
sfqi:
```
return &model.Book {
...
```

interactor/book.go
```diff
@@ -53,3 +69,51 @@ func (b *Book) Delete(*model.Book) error {
...
+	}
+
+	year := 0
+	var err error
+	yearString := yearRgx.FindString(book.Year)
+	if yearString != "" {
+		year, err = strconv.Atoi(yearString)
+		if err != nil {
+			fmt.Println("error while converting year from string to int", err)
+			return nil
```
sfqi: we should return this error instead of logging

interactor/book.go
```diff
@@ -53,3 +69,51 @@ func (b *Book) Delete(*model.Book) error {
...
+	if book.Identifier.Openlibrary != nil {
+		libraryId = book.Identifier.Openlibrary[0]
+	}
+	author := ""
+	if book.Author != nil {
+		author = book.Author[0].Name
+	}
+
+	year := 0
+	var err error
```
sfqi: do we need to declare these two in advance?

interactor/book.go
```diff
@@ -53,3 +69,51 @@ func (b *Book) Delete(*model.Book) error {
...
+	yearString := yearRgx.FindString(book.Year)
+	if yearString != "" {
```
sfqi: We try to avoid nested `if`s in Go, so probably better to do something like
```
if yearString == "" {
    return nil, fmt.Errorf(....) // or something like that
}

year, err := ...
```

interactor/book.go
```diff
@@ -53,3 +69,51 @@ func (b *Book) Delete(*model.Book) error {
 
 	return nil
 }
+
+func toBook(book *openlibrarydto.Book) (bm *model.Book) {
```
sfqi: no need to name the return param

interactor/book_test.go
```diff
@@ -1,8 +1,94 @@
...
+					ISBN13:      []string{"9780140447934"},
+					Openlibrary: []string{"OL7355422M"},
+				},
+				Author: []openlibrarydto.Author{
+					{Name: "Tolstoy"},
+				},
+				Cover: openlibrarydto.Cover{"https://covers.openlibrary.org/b/id/5049015-S.jpg"},
+				Year:  "2007",
+			},
+			nil)
```
sfqi:
```
    nil,
)
```

interactor/book_test.go
```diff
@@ -1,8 +1,94 @@
...
+		assert.Nil(book)
+		assert.Equal(err.Error(), clientError)
+	})
+	t.Run("Error creating book in database", func(t *testing.T) {
+		store := &repomock.Store{}
+		openlibClient := &openlibmock.Client{}
+		storeError := "Error saving book in database"
+
+		openlibClient.On("FetchBook", "0140447938").Return(&openlibrarydto.Book{Title: "War and Peace (Penguin Classics)"}, nil)
+		store.On("CreateBook", &model.Book{Title: "War and Peace (Penguin Classics)"}).Return(errors.New("Error saving book in database"))
```
sfqi: Just use `storeError` as the `Return` argument

interactor/book_test.go
```diff
@@ -1,8 +1,94 @@
...
 
 func TestBook_Create(t *testing.T) {
+	assert := assert.New(t)
+	//require := require.New(t)
+	t.Run("Error while fetching book", func(t *testing.T) {
+		store := &repomock.Store{}
+		openlibClient := &openlibmock.Client{}
+		clientError := "Error while fetching book: "
+
+		openlibClient.On("FetchBook", "0140447938222").Return(nil, errors.New("Error while fetching book: "))
```
sfqi: Just use `clientError` as the `Return` arg.

interactor/book_test.go
```diff
@@ -1,8 +1,94 @@
...
+	openlibrarydto "github.com/sfqi/library/openlibrary/dto"
+	openlibmock "github.com/sfqi/library/openlibrary/mock"
+	repomock "github.com/sfqi/library/repository/mock"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
 
 func TestBook_Create(t *testing.T) {
+	assert := assert.New(t)
+	//require := require.New(t)
```
sfqi: we can delete this, I guess

interactor/book.go
bokimilinkovic: Then we are explicit telling that we must send some Year, since yearString cant be empty ... I thought if we dont set some year value, to be set to 0 as default

sfqi: fair enough

PR #69, opened 2019-12-24
Feature/find by id interactor

StefanGajic:
https://trello.com/c/qV2fRdoP/99-add-book-interactor-find-by-id-method-with-tests

### DIFF
--- a/interactor/book.go
+++ b/interactor/book.go
@@ -20,6 +20,7 @@ type openlibraryClient interface {
 type Book struct {
 	store   store
 	openlib openlibraryClient
+	books   []model.Book
 }
 
 func NewBook(store store, olc openlibraryClient) *Book {
@@ -44,9 +45,10 @@ func (b *Book) Update(*model.Book) error {
 	return nil
 }
 
-func (b *Book) FindById(int) (*model.Book, error) {
+func (b *Book) FindById(id int) (*model.Book, error) {
 
-	return nil, nil
+	return b.store.FindBookById(id)
+	//2 scenarija, jedan kad se vrati dobro, drugi kad vrati gresku
 }
 
 func (b *Book) Delete(*model.Book) error {
--- a/interactor/book.go
+++ b/interactor/book.go
@@ -48,7 +48,6 @@ func (b *Book) Update(*model.Book) error {
 func (b *Book) FindById(id int) (*model.Book, error) {
 
 	return b.store.FindBookById(id)
-	//2 scenarija, jedan kad se vrati dobro, drugi kad vrati gresku
 }
 
 func (b *Book) Delete(*model.Book) error {
--- a/interactor/book_test.go
+++ b/interactor/book_test.go
@@ -55,18 +55,17 @@ func TestFindById(t *testing.T) {
 	})
 	t.Run("Cannot retrieve book", func(t *testing.T) {
 		store := &mock.Store{}
-
 		storeError := errors.New("Error finding ID from database")
 
 		b := interactor.NewBook(store, nil)
 
-		store.On("FindBookById", 12).Return(nil, errors.New("Error finding ID from database"))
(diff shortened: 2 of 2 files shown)


interactor/book_test.go
```diff
@@ -10,8 +19,40 @@ func TestBook_FindAll(t *testing.T) {
 
 }
 
-func TestBook_FindById(t *testing.T) {
+func TestFindById(t *testing.T) {
+	t.Run("Successfully retrieved book", func(t *testing.T) {
+		var db = &mock.Store{}
+		clmock := &olmock.Client{}
```
bokimilinkovic: do we really need mock Client here?

StefanGajic: its because of `NewBook()` which takes 2 arguments

bokimilinkovic: It can also take (db, nil)

interactor/book_test.go
```diff
@@ -10,8 +18,41 @@ func TestBook_FindAll(t *testing.T) {
...
+			Isbn:          "some isbn",
+			Isbn13:        "some isbon13",
+			OpenLibraryId: "again some id",
+			CoverId:       "some cover ID",
+			Year:          2019,
+		}, nil)
+
+		book, err := b.FindById(1)
+
+		assert.NotNil(t, book)
```
sfqi: Let's check what the exact return value is

interactor/book_test.go
```diff
@@ -10,8 +18,41 @@ func TestBook_FindAll(t *testing.T) {
 
 }
 
-func TestBook_FindById(t *testing.T) {
+func TestFindById(t *testing.T) {
+	t.Run("Successfully retrieved book", func(t *testing.T) {
+		var db = &mock.Store{}
```
sfqi: `store := &mock.Store{}`

interactor/book_test.go
```diff
@@ -10,8 +18,41 @@ func TestBook_FindAll(t *testing.T) {
...
+		}, nil)
+
+		book, err := b.FindById(1)
+
+		assert.NotNil(t, book)
+		assert.NoError(t, err)
+
+	})
+	t.Run("Cannot retrieve book", func(t *testing.T) {
+		var db = &mock.Store{}
```
sfqi: See above

interactor/book_test.go
```diff
@@ -10,8 +18,41 @@ func TestBook_FindAll(t *testing.T) {
...
+
+		b := interactor.NewBook(db, nil)
+
+		db.On("FindBookById", 12).Return(nil, errors.New("Error finding ID from database"))
+
+		book, err := b.FindById(12)
 
+		assert.Nil(t, book)
+		assert.Error(t, err)
```
sfqi: you could use the `assert := assert.New(t)` command

interactor/book_test.go
```diff
@@ -10,8 +18,41 @@ func TestBook_FindAll(t *testing.T) {
...
+		assert.NotNil(t, book)
+		assert.NoError(t, err)
+
+	})
+	t.Run("Cannot retrieve book", func(t *testing.T) {
+		var db = &mock.Store{}
+
+		b := interactor.NewBook(db, nil)
+
+		db.On("FindBookById", 12).Return(nil, errors.New("Error finding ID from database"))
```
sfqi: What I like to do is to previously initialize an error var
`storeError := errors.New("...")`
so that I can also check it in the assertion

interactor/book.go
```diff
@@ -20,6 +20,7 @@ type openlibraryClient interface {
 type Book struct {
 	store   store
 	openlib openlibraryClient
+	books   []model.Book
```
bokimilinkovic: I think we dont need this

sfqi: @bokimilinkovic You're right, we don't (I remember talking about it in person with someone)

PR #70, opened 2019-12-25
Added FindAll to interactor and TestFindAll also

StefanGajic:
https://trello.com/c/a8zfk0Da/101-add-book-interactor-find-all-method-with-tests

PR #72, opened 2019-12-25
Added Delete for interactor and TestDelete also

StefanGajic:
https://trello.com/c/YVn38GHH/102-add-book-interactor-delete-method-with-tests

### DIFF
--- a/interactor/book_test.go
+++ b/interactor/book_test.go
@@ -1,6 +1,14 @@
 package interactor_test
 
-import "testing"
+import (
+	"errors"
+	"testing"
+
+	"github.com/sfqi/library/domain/model"
+	"github.com/sfqi/library/interactor"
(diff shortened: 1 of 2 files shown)


interactor/book_test.go
```diff
@@ -14,8 +22,34 @@ func TestBook_FindById(t *testing.T) {
...
+		b := interactor.NewBook(db, nil)
+
+		bookToDelete := &model.Book{}
+
+		db.On("DeleteBook", bookToDelete).Return(errors.New("Error while deleting book"))
+
+		err := b.Delete(bookToDelete)
 
+		assert.Error(t, err)
+	})
```
sfqi: Take a look at the comments in e.g. https://github.com/sfqi/library/pull/70

Also no need to separate every code line with an empty one

PR #74, opened 2020-01-05
Used interactor in handler and bookloader tests

StefanGajic:
https://trello.com/c/3BEEEYIp/111-use-the-interactor-in-tests-for-handler-and-bookloader

### DIFF
--- a/handler/handler.go
+++ b/handler/handler.go
@@ -4,10 +4,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"regexp"
+
 	"github.com/sfqi/library/domain/model"
 	"github.com/sfqi/library/handler/dto"
-	"github.com/sfqi/library/interactor"
-	"regexp"
 
 	"net/http"
 )
@@ -15,7 +15,15 @@ import (
 var yearRgx = regexp.MustCompile(`[0-9]{4}`)
 
 type BookHandler struct {
-	Book *interactor.Book
+	Book bookInteractor
+}
+
+type bookInteractor interface {
+	FindAll() ([]*model.Book, error)
+	Create(bookRequest dto.CreateBookRequest) (*model.Book, error)
+	Update(book *model.Book, updateBookRequest dto.UpdateBookRequest) (*model.Book, error)
+	FindById(id int) (*model.Book, error)
+	Delete(book *model.Book) error
 }
 
 func toBookResponse(b model.Book) *dto.BookResponse {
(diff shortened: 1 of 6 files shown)


handler/handler.go
```diff
@@ -5,34 +5,25 @@ import (
...
 }
 
-type openLibraryClient interface {
-	FetchBook(isbn string) (*openlibrarydto.Book, error)
+type bookInteractor interface {
+	FindAll() ([]*model.Book, error)
+	Create(bookRequest dto.CreateBookRequest) (*model.Book, error)
+	Update(book *model.Book, updateBookRequest dto.UpdateBookRequest) (*model.Book, error)
+	FindById(id int) (*model.Book, error)
+	Delete(book *model.Book) error
```
bokimilinkovic: why creating this interface here when we have struct Book in package interactor? 
https://github.com/sfqi/library/blob/chore/adding_interactor_to_handler/interactor/book.go

sfqi: @bokimilinkovic Are you objecting to the interface itself or to the fact that it's defined in this file (or package)?

bokimilinkovic: I am objecting to interface.. I think that we don't need this here, we have Book(from interactor package) with all methods attached to it

sfqi: Not sure why the new interface and `interactor.Book` would be in conflict - `interactor.Book` implements the interface. It's a dependency that we want to be able to mock.

middleware/bookloader.go
```diff
@@ -3,15 +3,20 @@ package middleware
...
-	"github.com/sfqi/library/interactor"
 	"strconv"
 
 	"github.com/gorilla/mux"
+	"github.com/sfqi/library/domain/model"
+
 	"net/http"
 )
 
+type BookInteractor interface {
```
sfqi: Make it unexported

handler/handler_test.go
```diff
@@ -166,17 +162,21 @@ func TestUpdate(t *testing.T) {
...
-			Year:          2019,
-		}).Return(errors.New("error updating book"))
-		bookHandler.Db = db
+			Year:          2000,
+		}, dto.UpdateBookRequest{
+			Title: "test title",
+			Year:  2019,
+		},
+		).Return(nil, errors.New("error updating book"))
+		bookHandler.Book = interactor
```
sfqi: Could you also rename the field so that it's `bookHandler.Interactor = interactor`?

PR #75, opened 2020-01-06
Defined HTTPError type in the handler

StefanGajic:
https://trello.com/c/EbNspkFc/109-define-httperror-type-in-the-handler-pkg

### DIFF
--- a/handler/handler.go
+++ b/handler/handler.go
@@ -18,6 +18,24 @@ import (
 
 var yearRgx = regexp.MustCompile(`[0-9]{4}`)
 
+type HTTPError struct {
+	code     int
+	internal error
+	context  string
+}
+
+func (h HTTPError) Error() string {
+	return fmt.Sprintf("HTTP %d: %s", h.code, h.context)
+}
+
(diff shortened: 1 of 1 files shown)


handler/handler.go
```diff
@@ -18,6 +18,24 @@ import (
...
 var yearRgx = regexp.MustCompile(`[0-9]{4}`)
 
+type HTTPError struct {
+	code     int
+	internal error
+	context  string
+}
+
+func (h HTTPError) Error() string {
+	return fmt.Sprintf("HTTP %d: %s", h.code, h.context)
```
sfqi: You want to print the internal error as well

handler/handler.go
```diff
@@ -18,6 +18,24 @@ import (
...
+}
+
+func newHTTPError(code int, err error, ctx string) *HTTPError {
```
sfqi: The context should be optional, so we should be able to do something like:
`return newHTTPError(err, http.BadStatusRequest).Wrap("optional context")`

sfqi:
Ahem: `.Wrap("optional context")`

PR #76, opened 2020-01-06
initial custom handler returning error

bokimilinkovic:
https://trello.com/c/2eG7JvRf/110-modify-book-handler-methods-so-that-they-return-an-error


handler/customhandler.go
```diff
@@ -0,0 +1,23 @@
...
+
+type ErrorHandler struct {
+	Logger *logrus.Logger
+}
+
+func (eh *ErrorHandler) Wrap(handler customHandler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		eh.Logger.Println(r.URL.Path + r.Method)
+		handler.ServeHTTP(w, r)
+	})
```
bokimilinkovic: I dont know if you meant something like this, I am not sure how to log error if handler returns one ... Does this method:
func (ch *customHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)   needs to return also error so i can log it?

handler/customhandler.go
```diff
@@ -0,0 +1,23 @@
...
+
+}
+
+type ErrorHandler struct {
+	Logger *logrus.Logger
+}
+
+func (eh *ErrorHandler) Wrap(handler customHandler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		eh.Logger.Println(r.URL.Path + r.Method)
```
sfqi: We don't need this (it's already done by middleware

handler/customhandler.go
sfqi: Could something like this work:

Since `handler` is a function
```
err := handler(w, r)
eh.Logger.Error(err)
```

or maybe even just `eh.Logger.Error(handler(w, r))`

?

bokimilinkovic: yes handler is type of function, but i need to call it somehow in middleware .. handler(w,r) is not working, only recomendation for that is handler.ServeHTTP,(w,r) so i guess i need to modify ServeHTTP so it returns an error

bokimilinkovic: func (ch *customHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) error {

}
But then i have no idea, what to return here since it is expecting error

main.go
```diff
@@ -52,17 +52,19 @@ func main() {
 		Logger: log.New(),
 	}
 
+	handleFunc := handler.ErrorHandler{Logger: log.New()}
```
sfqi: Why not define a single logger object and inject it where needed?

sfqi: how about:
`handleFunc := handler.ErrorHandler{Logger: logger}.Wrap`

Then we could do:
`r.Handle("/books", handleFunc(bookHandler.Index)).Methods("GET")`

handler/error.go
```diff
@@ -0,0 +1,62 @@
...
+
+type ErrorHandler struct {
+	Logger *logrus.Logger
+}
+
+func (eh *ErrorHandler) Wrap(handler customHandler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var errorMsg string
+		err := handler(w, r)
+		eh.Logger.Error(err)
```
sfqi: This should be within the `if` branch below (we don't want to log if the error is nil)

handler/book.go
```diff
@@ -14,36 +14,6 @@ import (
```
sfqi: @bokimilinkovic I thought you wanted to name the file `bookhandler.go`?

bokimilinkovic: yeah, but I first renamed it before i asked you to rename in bookhandler, so I forgot. Stefan will now fix this things , since he has to do testing of wrap method

PR #77, opened 2020-01-08
Added middleware.Bookloader instead of middleware.Bookinteractor to f…

StefanGajic:
https://trello.com/c/jnRQMYrT/112-fix-bookloader-test


middleware/bookloader_test.go
```diff
@@ -15,7 +15,6 @@ import (
 
 type bookHandler struct {
 	bookFromContext *model.Book
```
sfqi: Can we remove this field as well?

PR #78, opened 2020-01-09
Feature/handling erros from handlers

bokimilinkovic:
https://trello.com/c/OEvxLkaH/113-handle-errors-returned-by-handlers


handler/handler.go
```diff
@@ -71,13 +63,13 @@ func toBookResponse(b model.Book) *dto.BookResponse {
...
 
-func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) {
+func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) *HTTPError {
 
 	w.Header().Set("Content-Type", "application/json")
 	allBooks, err := b.Interactor.FindAll()
 	if err != nil {
-		http.Error(w, "Error finding books", http.StatusInternalServerError)
-		return
+		//http.Error(w, "Error finding books", http.StatusInternalServerError)
```
StefanGajic: Remove all the comments

handler/customhandler.go
```diff
@@ -0,0 +1,26 @@
...
+type ErrorHandler struct {
+	Logger *logrus.Logger
+}
+
+func (eh *ErrorHandler) Wrap(handler customHandler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		err := handler(w, r)
+		if err != nil {
+			w.WriteHeader(err.code)
+			eh.Logger.Error(err) //1. logging
```
StefanGajic: Remove the comment

handler/customhandler.go
```diff
@@ -14,6 +15,12 @@ type ErrorHandler struct {
 func (eh *ErrorHandler) Wrap(handler customHandler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		err := handler(w, r)
-		eh.Logger.Error(err)
+		if err != nil {
+			w.WriteHeader(err.code)
+			eh.Logger.Error(err)
+			if err.code < http.StatusInternalServerError {
+				fmt.Fprintf(w, err.Error())
```
sfqi: I think I get why you didn't use `http.Error`, but the problem here is that `Fprintf` returns an error, which needs to be checked and handled as well.

Maybe we should have a switch and use `http.Error` in both cases

handler/handler.go
```diff
@@ -63,13 +63,12 @@ func toBookResponse(b model.Book) *dto.BookResponse {
...
 
-func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
+func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) *HTTPError {
 
 	w.Header().Set("Content-Type", "application/json")
 	allBooks, err := b.Interactor.FindAll()
 	if err != nil {
-		http.Error(w, "Error finding books", http.StatusInternalServerError)
-		return err
+		return newHTTPError(http.StatusInternalServerError, errors.New("Error finding books"))
```
sfqi: Why are we creating a new error when there's already a perfectly informative one (`err`)?

handler/handler.go
```diff
@@ -80,117 +79,90 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
 
 	err = json.NewEncoder(w).Encode(bookResponses)
 	if err != nil {
-		fmt.Println("error while getting books: ", err)
-		http.Error(w, "Bad request", http.StatusBadRequest)
-		return err
+		return newHTTPError(http.StatusBadRequest, errors.New("Bad request"))
```
sfqi: Again, why not use `err`?

handler/handler.go
```diff
@@ -80,117 +79,90 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
...
-func (b *BookHandler) Create(w http.ResponseWriter, r *http.Request) error {
+func (b *BookHandler) Create(w http.ResponseWriter, r *http.Request) *HTTPError {
 	w.Header().Set("Content-Type", "application/json")
 
 	var createBook dto.CreateBookRequest
 
 	if err := json.NewDecoder(r.Body).Decode(&createBook); err != nil {
-		errorDecodingBook(w, err)
-		return err
+		return newHTTPError(http.StatusBadRequest, errors.New("Error while decoding from request body"))
```
sfqi: `err`?

handler/handler.go
```diff
@@ -80,117 +79,90 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
...
 	}
 
 	fmt.Println(createBook.ISBN)
 	book, err := b.Interactor.Create(createBook)
 
 	if err != nil {
-		fmt.Println("error while fetching book: ", err)
-		http.Error(w, "Error while fetching book: "+err.Error(), http.StatusInternalServerError)
-		return err
+		return newHTTPError(http.StatusInternalServerError, errors.New("Internal server error: "+err.Error()))
```
sfqi: `err`

handler/handler.go
```diff
@@ -80,117 +79,90 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
...
+
```
sfqi: do we need this empty line?

handler/handler.go
```diff
@@ -80,117 +79,90 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
...
 	}
 
 	bookResponse := *toBookResponse(*book)
+
```
sfqi: do we need this empty line?

handler/handler.go
```diff
@@ -80,117 +79,90 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
...
 	if err := json.NewEncoder(w).Encode(bookResponse); err != nil {
-		errorEncoding(w, err)
-		return err
+		return newHTTPError(http.StatusBadRequest, errors.New("Bad request"))
```
sfqi: This one should be internal (and also: use `err`)

handler/handler.go
```diff
@@ -80,117 +79,90 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
...
 
-func (b *BookHandler) Update(w http.ResponseWriter, r *http.Request) error {
+func (b *BookHandler) Update(w http.ResponseWriter, r *http.Request) *HTTPError {
 
 	w.Header().Set("Content-Type", "application/json")
 	book, ok := r.Context().Value("book").(*model.Book)
 	if !ok {
-		errorContex(w, errors.New("error retrieving book from context"))
-		return errors.New("Error retrieving book from context")
+		return newHTTPError(http.StatusInternalServerError, errors.New("Internal server error: Can not return book from context"))
```
sfqi:
```
err := errors.New("error retrieving book from context")
return newHTTPError(http.StatusInternalServerError, err)
```

handler/handler.go
```diff
@@ -80,117 +79,90 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
...
 	}
 	updateBookRequest := dto.UpdateBookRequest{}
 	err := json.NewDecoder(r.Body).Decode(&updateBookRequest)
 	if err != nil {
-		errorDecodingBook(w, err)
-		return err
+		return newHTTPError(http.StatusBadRequest, errors.New("Error while decoding from request body"))
```
sfqi: `err`

handler/handler.go
```diff
@@ -80,117 +79,90 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
...
 	}
 
 	updatedBook, err := b.Interactor.Update(book, updateBookRequest)
 	if err != nil {
-		http.Error(w, "error updating book", http.StatusInternalServerError)
-		return err
+		return newHTTPError(http.StatusInternalServerError, errors.New("error updating book"))
```
sfqi: `err`

handler/handler.go
```diff
@@ -80,117 +79,90 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
...
 	}
 	bookResponse := *toBookResponse(*updatedBook)
 
 	err = json.NewEncoder(w).Encode(bookResponse)
 	if err != nil {
-		errorEncoding(w, err)
-		return err
+		return newHTTPError(http.StatusBadRequest, errors.New("Bad request"))
```
sfqi: Internal and use `err`

handler/handler.go
```diff
@@ -80,117 +79,90 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
...
-func (b *BookHandler) Get(w http.ResponseWriter, r *http.Request) error {
+func (b *BookHandler) Get(w http.ResponseWriter, r *http.Request) *HTTPError {
 
 	w.Header().Set("Content-Type", "application/json")
 
 	book, ok := r.Context().Value("book").(*model.Book)
 	if !ok {
-		errorContex(w, errors.New("error retrieving book from context"))
-		return errors.New("error retrieving book from context")
+		return newHTTPError(http.StatusInternalServerError, errors.New("Internal server error: Can not return book from context"))
```
sfqi: `err`

handler/handler.go
```diff
@@ -80,117 +79,90 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
...
 	}
 
 	bookResponse := *toBookResponse(*book)
 
 	if err := json.NewEncoder(w).Encode(bookResponse); err != nil {
-		errorEncoding(w, err)
-		return err
+		return newHTTPError(http.StatusBadRequest, errors.New("Bad request"))
```
sfqi: Internal & use `err`

handler/handler.go
```diff
@@ -80,117 +79,90 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
...
 }
 
-func (b *BookHandler) Delete(w http.ResponseWriter, r *http.Request) error {
+func (b *BookHandler) Delete(w http.ResponseWriter, r *http.Request) *HTTPError {
 	w.Header().Set("Content-Type", "application/json")
 	book, ok := r.Context().Value("book").(*model.Book)
 	if !ok {
-		errorContex(w, errors.New("error retrieving book from context"))
-		return errors.New("error retrieving book from context")
+		return newHTTPError(http.StatusInternalServerError, errors.New("Internal server error: Can not return book from context"))
```
sfqi: `err`

handler/handler.go
```diff
@@ -80,117 +79,90 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) error {
...
 	}
 	fmt.Println("Book from context: ", book)
 
 	if err := b.Interactor.Delete(book); err != nil {
-		http.Error(w, "Error while deleting book", http.StatusInternalServerError)
-		return err
+		return newHTTPError(http.StatusInternalServerError, errors.New("Error while deleting book"))
```
sfqi: `err`

handler/customhandler.go
```diff
@@ -1,11 +1,12 @@
 package handler
```
sfqi: let's call this file `error.go`

handler/handler.go
```diff
@@ -15,18 +15,18 @@ import (
...
+	return fmt.Sprintf("HTTP %d: %s. Error: %s", h.Code, h.context, h.internal)
 }
 
 func newHTTPError(code int, err error) *HTTPError {
 	return &HTTPError{
-		code:     code,
+		Code:     code,
 		internal: err,
 	}
 }
```
sfqi: Let's move error stuff to error.go

handler/handler.go
```diff
@@ -15,18 +15,18 @@ import (
 var yearRgx = regexp.MustCompile(`[0-9]{4}`)
```
sfqi: let's rename this file to book.go

handler/customhandler.go
```diff
@@ -14,6 +15,12 @@ type ErrorHandler struct {
 func (eh *ErrorHandler) Wrap(handler customHandler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		err := handler(w, r)
-		eh.Logger.Error(err)
+		if err != nil {
+			w.WriteHeader(err.Code)
+			eh.Logger.Error(err)
+			if err.Code < http.StatusInternalServerError {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
```
sfqi: 1. we don't want the status to be internal server error here. Just use the error code

2. when you so it this way, you'll write to header twice, I think

I was thinking about something like this:
```
var errorMsg string
switch err.Code{
case <something>:
   // write or don't write to errorMsg
case <something else>:
   // write or don't write to errorMsg
}
http.Error(...
```

handler/handler.go
```diff
@@ -15,18 +15,18 @@ import (
...
 
 type HTTPError struct {
-	code     int
+	Code     int
```
sfqi: I'm guessing this is exported because you need it in the tests?
Ideally we would compare the entire error, not just the code, but the internal field is probably difficult to deal with. Then let's add a `Code()` method and keep the field unexported.

handler/handler_test.go
```diff
@@ -79,8 +79,9 @@ func TestIndex(t *testing.T) {
 			}}, nil)
 
 		bookHandler.Interactor = interactor
-		handler := http.HandlerFunc(bookHandler.Index)
-		handler.ServeHTTP(rr, req)
+		handler := bookHandler.Index
+		httpError := handler(rr, req)
+		assert.Nil(httpError)
 		require.Equal(http.StatusOK, rr.Code)
```
sfqi: This should be an `assert`, I think

handler/handler_test.go
```diff
@@ -79,8 +79,9 @@ func TestIndex(t *testing.T) {
 			}}, nil)
 
 		bookHandler.Interactor = interactor
-		handler := http.HandlerFunc(bookHandler.Index)
-		handler.ServeHTTP(rr, req)
+		handler := bookHandler.Index
+		httpError := handler(rr, req)
```
sfqi: Could we simplify these two lines and just do `err := bookHandler.Index(rr, req)`?
The question applies to pretty much every test

handler/handler.go
bokimilinkovic: Or maybe bookhandler.go ? Since we have 5 files named book.go

sfqi: ok, let's try it

handler/error_test.go
```diff
@@ -0,0 +1,55 @@
...
+				context:  "with some context",
+			},
+			message: "HTTP 500: with some context Some error occured with code 500",
+		},
+	}
+	for _, tt := range tests {
+		t.Run("test", func(t *testing.T) {
+			errMsg := tt.error.Error()
+			if tt.message != errMsg {
+				t.Errorf("expected %q; got %q", tt.message, errMsg)
```
sfqi: why not use `assert`?

handler/error_test.go
```diff
@@ -0,0 +1,55 @@
...
+			error: HTTPError{
+				code:     500,
+				internal: errors.New("Some error occured with code 500"),
+				context:  "with some context",
+			},
+			message: "HTTP 500: with some context Some error occured with code 500",
+		},
+	}
+	for _, tt := range tests {
+		t.Run("test", func(t *testing.T) {
```
sfqi: What you could do is add a `name` field to the test struct, so that tests actually have different names

handler/error_test.go
sfqi: once these to comments are implemented, you can merge the PR

PR #79, opened 2020-01-13
Feature/handling erros from handlers

bokimilinkovic:
https://trello.com/c/nAcxrW6f/115-test-errorhandler


handler/error_test.go
```diff
@@ -68,21 +69,63 @@ func testHandler(w http.ResponseWriter, r *http.Request) *HTTPError {
...
+		logger := logrus.New()
+		rr := httptest.NewRecorder()
+		errHandler := ErrorHandler{logger}
+		newHandler := errHandler.Wrap(testHandler2)
+		newHandler.ServeHTTP(rr, req)
+		assert.Equal(http.StatusInternalServerError, rr.Code)
+		expected := "\n"
+		assert.Equal(expected, rr.Body.String())
+	})
+	t.Run("Status code 500", func(t *testing.T) {
```
bokimilinkovic: Status code 400***

handler/error.go
```diff
@@ -46,13 +46,11 @@ func (eh *ErrorHandler) Wrap(handler customHandler) http.Handler {
...
 		err := handler(w, r)
-		eh.Logger.Error(err)
 
 		if err != nil {
-			switch err.code {
-			case http.StatusInternalServerError:
+			eh.Logger.Error(err)
 
-			default:
+			if err.code < 500 {
```
sfqi: I was going to do this myself because I was the one who initially suggested using a switch :)

just use `http.StatusInternalServerError` instead of `500`

handler/error_test.go
```diff
@@ -58,3 +63,69 @@ func TestError(t *testing.T) {
...
+	}
+}
+
+func TestWrap(t *testing.T) {
+	assert := assert.New(t)
+	t.Run("Test Wrap", func(t *testing.T) {
+		req, err := http.NewRequest("GET", "/testerrorhandler", nil)
+		if err != nil {
+			t.Fatal(err)
+		}
```
sfqi: use `require` instead

handler/error_test.go
```diff
@@ -58,3 +63,69 @@ func TestError(t *testing.T) {
...
+		logger := logrus.New()
+		rr := httptest.NewRecorder()
+		errHandler := ErrorHandler{logger}
+		newHandler := errHandler.Wrap(testHandler)
+		newHandler.ServeHTTP(rr, req)
```
sfqi: How about `ErrorHandler{logger}.Wrap(testHandler).ServeHTTP(rr, req)`?

handler/error_test.go
```diff
@@ -58,3 +63,69 @@ func TestError(t *testing.T) {
...
+		assert.Equal(http.StatusOK, rr.Code)
+		fmt.Println(rr.Code)
```
sfqi: No need to print

handler/error_test.go
```diff
@@ -58,3 +63,69 @@ func TestError(t *testing.T) {
...
+	})
+	t.Run("Status code 500", func(t *testing.T) {
+		req, err := http.NewRequest("GET", "/testerrorhandler2", nil)
+		if err != nil {
+			t.Fatal(err)
+		}
```
sfqi: `require`

handler/error_test.go
```diff
@@ -58,3 +63,69 @@ func TestError(t *testing.T) {
...
+		logger := logrus.New()
+		rr := httptest.NewRecorder()
+		errHandler := ErrorHandler{logger}
+		newHandler := errHandler.Wrap(testHandler2)
+		newHandler.ServeHTTP(rr, req)
```
sfqi: Again, maybe we could one-line this.

handler/error_test.go
```diff
@@ -58,3 +63,69 @@ func TestError(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		logger := logrus.New()
+		rr := httptest.NewRecorder()
+		errHandler := ErrorHandler{logger}
+		newHandler := errHandler.Wrap(testHandler3)
+		newHandler.ServeHTTP(rr, req)
+		assert.Equal(http.StatusBadRequest, rr.Code)
+		expected := "HTTP 400:  with some context error with status code 400\n"
```
sfqi: hm, I'd like there to be a `:` after the context message - HTTP 400: with some context: error with status...

also, is there an extra white space after `HTTP 400:`?

handler/error_test.go
```diff
@@ -58,3 +63,69 @@ func TestError(t *testing.T) {
...
+	t.Run("Status code 500", func(t *testing.T) {
+		req, err := http.NewRequest("GET", "/testerrorhandler3", nil)
+		if err != nil {
+			t.Fatal(err)
+		}
+		logger := logrus.New()
+		rr := httptest.NewRecorder()
+		errHandler := ErrorHandler{logger}
+		newHandler := errHandler.Wrap(testHandler3)
+		newHandler.ServeHTTP(rr, req)
```
sfqi: Same comments as for the tests above

handler/error_test.go
```diff
@@ -58,3 +63,69 @@ func TestError(t *testing.T) {
...
+}
+
+func testHandler2(w http.ResponseWriter, r *http.Request) *HTTPError {
+	return &HTTPError{
+		code:     500,
+		internal: errors.New("error with status code 500"),
+		context:  "",
+	}
+}
+func testHandler3(w http.ResponseWriter, r *http.Request) *HTTPError {
```
sfqi: How about `handler` (or `okHandler`), `badRequest` handler, `internalServerErrorHandler`?

handler/error_test.go
```diff
@@ -58,3 +63,69 @@ func TestError(t *testing.T) {
...
+		code:     400,
+		internal: errors.New("error with status code 400"),
+		context:  " with some context",
+	}
+}
+
+func TestWrap(t *testing.T) {
+	assert := assert.New(t)
+	t.Run("Test Wrap", func(t *testing.T) {
+		req, err := http.NewRequest("GET", "/testerrorhandler", nil)
```
sfqi: `"/testerrorhandler" -> "/"` is fine, I think. The path shouldn't really matter

handler/error_test.go
```diff
@@ -58,3 +63,69 @@ func TestError(t *testing.T) {
...
+		logger := logrus.New()
+		rr := httptest.NewRecorder()
+		errHandler := ErrorHandler{logger}
+		newHandler := errHandler.Wrap(testHandler)
+		newHandler.ServeHTTP(rr, req)
+		assert.Equal(http.StatusOK, rr.Code)
+		fmt.Println(rr.Code)
+	})
+	t.Run("Status code 500", func(t *testing.T) {
+		req, err := http.NewRequest("GET", "/testerrorhandler2", nil)
```
sfqi: `"/"`

main.go
```diff
@@ -2,13 +2,14 @@ package main
```
sfqi: Changes in this file seem to belong to another task (no need to remove them now, just noticing)

handler/error_test.go
bokimilinkovic: yeah, good idea

handler/error_test.go
```diff
@@ -58,3 +63,60 @@ func TestError(t *testing.T) {
...
+
+func TestWrap(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+	t.Run("Test Wrap", func(t *testing.T) {
+		req, err := http.NewRequest("GET", "/", nil)
+		require.NoError(err)
+		logger := logrus.New()
+		rr := httptest.NewRecorder()
+		(&ErrorHandler{logger}).Wrap(okHandler).ServeHTTP(rr, req)
```
sfqi: do we have to use a pointer to `ErrorHandler` here? (genuinely asking)

handler/error_test.go
```diff
@@ -58,3 +63,60 @@ func TestError(t *testing.T) {
...
+	})
+	t.Run("Status code 400", func(t *testing.T) {
+		req, err := http.NewRequest("GET", "/", nil)
+		require.NoError(err)
+		logger := logrus.New()
+		rr := httptest.NewRecorder()
+		(&ErrorHandler{logger}).Wrap(badRequest).ServeHTTP(rr, req)
+
+		assert.Equal(http.StatusBadRequest, rr.Code)
+		expected := "HTTP 400: with some context error with status code 400\n"
```
sfqi: quoting myself: "hm, I'd like there to be a : after the context message - with some contex**t: e**rror with status..."

handler/error_test.go
```diff
@@ -58,3 +63,60 @@ func TestError(t *testing.T) {
...
+}
+
+func internalServerErrorHandler(w http.ResponseWriter, r *http.Request) *HTTPError {
+	return &HTTPError{
+		code:     500,
+		internal: errors.New("error with status code 500"),
+		context:  "",
+	}
+}
+func badRequest(w http.ResponseWriter, r *http.Request) *HTTPError {
```
sfqi: `badRequestHandler`, no?

handler/error_test.go
bokimilinkovic: i guess we dont

StefanGajic: it works without it but you told us it is always a good practice to use pointers

sfqi: Yeah, but we're not passing it as an argument or anything (it's not even assigned to a var). We just call a method on it in one line.

PR #80, opened 2020-01-13
Created migration for loans, also modified Id to ID in loan.go

bokimilinkovic:
https://trello.com/c/nAcxrW6f/115-test-errorhandler

### DIFF
--- a/domain/model/loan.go
+++ b/domain/model/loan.go
@@ -31,7 +31,6 @@ func NewLoan(userId int, bookId int, state loanType) (*Loan, error) {
 		UserID:        userId,
 		BookID:        bookId,
 		Type:          state,
-		CreatedAt:     time.Now(),
 	}, nil
 }
 
--- a/domain/model/loan.go
+++ b/domain/model/loan.go
@@ -21,10 +21,10 @@ type Loan struct {
 	CreatedAt     time.Time
 }
 
-func NewLoan(userId int, bookId int, state loanType) *Loan {
+func NewLoan(userId int, bookId int, state loanType) (*Loan, error) {
 	uuid, err := uuid.NewUUID()
 	if err != nil {
-		return nil
+		return nil, err
 	}
 	return &Loan{
 		TransactionID: uuid.String(),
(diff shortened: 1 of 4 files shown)


domain/model/loan.go
```diff
@@ -1,44 +1,44 @@
...
 }
 
-func NewLoan(userId int, bookId int, state loanType) *Loan{
-	uuid ,err := uuid.NewUUID()
-	if err != nil{
-		fmt.Println(err)
+func NewLoan(userId int, bookId int, state loanType) *Loan {
+	uuid, err := uuid.NewUUID()
+	if err != nil {
+		return nil
```
sfqi: got to return the error as well

migrations/4_create_loan_table.up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE loans(
+    id serial primary key,
+    transaction_id varchar(255) unique,
```
sfqi: not unique, every borrowed-returned pair will have the same transaction id

migrations/4_create_loan_table.up.sql
```diff
@@ -0,0 +1,9 @@
...
+    user_id int not null,
+    book_id int unique not null,
```
sfqi: not unique, the same book can be loaned multiple times

domain/model/loan.go
```diff
@@ -1,44 +1,44 @@
...
 		TransactionID: uuid.String(),
-		UserID:userId,
-		BookID:bookId,
-		Type:state,
-		CreatedAt : time.Now(),
-	}
+		UserID:        userId,
+		BookID:        bookId,
+		Type:          state,
+		CreatedAt:     time.Now(),
```
sfqi: I think gorm sets the value of `CreatedAt`. If so, we don't have to do it here. Can you double-check?

bokimilinkovic: yes, you are right, it sets automatically

sfqi: probably fine if you remove it, then

PR #85, opened 2020-01-14
Feature/create loan interactor

StefanGajic:
https://trello.com/c/ec7qNe4w/122-add-create-loan-interactor-method

### DIFF
--- a/domain/model/loan.go
+++ b/domain/model/loan.go
@@ -1,7 +1,6 @@
 package model
 
 import (
-	"github.com/google/uuid"
 	"time"
 )
 
@@ -22,16 +21,10 @@ type Loan struct {
 }
 
 func NewLoan(userId int, bookId int, state LoanType) (*Loan, error) {
-	uuid, err := uuid.NewUUID()
-	if err != nil {
-		return nil, err
-	}
-
 	return &Loan{
-		TransactionID: uuid.String(),
-		UserID:        userId,
-		BookID:        bookId,
-		Type:          state,
+		UserID: userId,
+		BookID: bookId,
+		Type:   state,
 	}, nil
 }
 
--- a/domain/model/loan.go
+++ b/domain/model/loan.go
@@ -20,18 +20,19 @@ type Loan struct {
 	CreatedAt     time.Time
 }
 
-func NewLoan(userId int, bookId int, state LoanType) (*Loan, error) {
+func NewLoan(userId int, bookId int, uuid string, state LoanType) (*Loan, error) {
 	return &Loan{
-		UserID: userId,
-		BookID: bookId,
-		Type:   state,
+		UserID:        userId,
+		BookID:        bookId,
+		TransactionID: uuid,
+		Type:          state,
 	}, nil
 }
 
-func ReturnedLoan(userid int, bookid int) (*Loan, error) {
-	return NewLoan(userid, bookid, returned)
+func ReturnedLoan(userid int, bookid int, uuid string) (*Loan, error) {
+	return NewLoan(userid, bookid, uuid, returned)
 }
 
-func BorrowedLoan(userid int, bookid int) (*Loan, error) {
-	return NewLoan(userid, bookid, borrowed)
+func BorrowedLoan(userid int, bookid int, uuid string) (*Loan, error) {
+	return NewLoan(userid, bookid, uuid, borrowed)
 }
--- a/domain/model/loan.go
+++ b/domain/model/loan.go
@@ -4,11 +4,11 @@ import (
 	"time"
 )
 
-type LoanType int
+type loanType int
 
 const (
-	borrowed LoanType = 0
-	returned LoanType = 1
+	borrowed loanType = 0
+	returned loanType = 1
 )
 
 type Loan struct {
@@ -16,16 +16,16 @@ type Loan struct {
 	TransactionID string
 	UserID        int
 	BookID        int
-	Type          LoanType
+	Type          loanType
 	CreatedAt     time.Time
 }
 
-func NewLoan(userId int, bookId int, uuid string, state LoanType) (*Loan, error) {
+func NewLoan(userId int, bookId int, uuid string, loanType loanType) (*Loan, error) {
 	return &Loan{
 		UserID:        userId,
 		BookID:        bookId,
 		TransactionID: uuid,
-		Type:          state,
+		Type:          loanType,
 	}, nil
 }
 
--- a/domain/model/loan.go
+++ b/domain/model/loan.go
@@ -5,11 +5,11 @@ import (
 	"time"
 )
 
-type loanType int
+type LoanType int
 
 const (
-	borrowed loanType = 0
-	returned loanType = 1
+	borrowed LoanType = 0
+	returned LoanType = 1
 )
 
 type Loan struct {
@@ -17,15 +17,16 @@ type Loan struct {
 	TransactionID string
 	UserID        int
 	BookID        int
-	Type          loanType
+	Type          LoanType
 	CreatedAt     time.Time
 }
 
-func NewLoan(userId int, bookId int, state loanType) (*Loan, error) {
+func NewLoan(userId int, bookId int, state LoanType) (*Loan, error) {
 	uuid, err := uuid.NewUUID()
 	if err != nil {
 		return nil, err
 	}
+
 	return &Loan{
 		TransactionID: uuid.String(),
 		UserID:        userId,
--- a/domain/model/loan.go
+++ b/domain/model/loan.go
@@ -20,19 +20,19 @@ type Loan struct {
 	CreatedAt     time.Time
 }
 
-func NewLoan(userId int, bookId int, uuid string, loanType loanType) (*Loan, error) {
+func newLoan(userId int, bookId int, uuid string, loanType loanType) *Loan {
 	return &Loan{
 		UserID:        userId,
 		BookID:        bookId,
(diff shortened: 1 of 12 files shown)


interactor/generator.go
```diff
@@ -0,0 +1,14 @@
...
+type generator struct {
+}
+
+func (g *generator) GenerateUUID() (string, error) {
+	uuid, err := uuid.NewUUID()
+	if err != nil {
+		return "", err
+	}
+	return uuid.String(), nil
+}
```
sfqi: how about we put this into `service/uuid`?

and use it e.g. like this: `uuid.Generator{}.Do()`?

interactor/generator.go
```diff
@@ -0,0 +1,14 @@
+package interactor
+
+import "github.com/google/uuid"
+
+type generator struct {
+}
```
sfqi: `type generator struct{}`

interactor/loans.go
```diff
@@ -9,15 +9,22 @@ type loansStore interface {
 	FindAllLoans() ([]*model.Loan, error)
 	FindLoansByBookID(int) ([]*model.Loan, error)
 	FindLoansByUserID(int) ([]*model.Loan, error)
+	CreateLoan(*model.Loan) error
+}
+
+type loansGenerator interface {
+	GenerateUUID() (string, error)
 }
```
sfqi: how about 
```
type uuidGenerator interface {
    Do() (string, error)
}
```

interactor/loans.go
```diff
@@ -30,3 +37,13 @@ func (l *Loan) FindAll() ([]*model.Loan, error) {
 
 	return l.loansStore.FindAllLoans()
 }
+
+func (l *Loan) CreateLoan(userID int, bookID int, state model.LoanType) error {
+	loan, err := model.NewLoan(userID, bookID, state)
```
sfqi: what do you think about passing the uuid as an argument here?

also why are we calling the type argument `state`?

sfqi:
@bokimilinkovic did you push some changes? I don't see any new commits

service/uuid.go
```diff
@@ -0,0 +1,13 @@
...
+import "github.com/google/uuid"
+
+type Generator struct{}
+
+func (g *Generator) Do() (string, error) {
+	uuid, err := uuid.NewUUID()
+	if err != nil {
+		return "", err
+	}
+	return uuid.String(), nil
```
sfqi: hm, maybe we could test this

service/uuid.go
```diff
@@ -0,0 +1,13 @@
...
+
+import "github.com/google/uuid"
+
+type Generator struct{}
+
+func (g *Generator) Do() (string, error) {
+	uuid, err := uuid.NewUUID()
+	if err != nil {
+		return "", err
+	}
```
sfqi: add an empty line after the `}`

service/mock/uuid.go
```diff
@@ -0,0 +1,18 @@
...
+
+type Generator struct {
+	mock.Mock
+}
+
+func (g *Generator) Do() (string, error) {
+	args := g.Called()
+	if args.Get(0) != "" {
+		return fmt.Sprintf("%v", args.Get(0)), nil
+	}
```
sfqi: add an empty line after the `}`

service/mock/uuid.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+type Generator struct {
+	mock.Mock
+}
+
+func (g *Generator) Do() (string, error) {
+	args := g.Called()
+	if args.Get(0) != "" {
+		return fmt.Sprintf("%v", args.Get(0)), nil
```
sfqi: isn't `args.String(0)` or something the usual way?

repository/mock/store.go
```diff
@@ -72,3 +72,11 @@ func (s *Store) FindLoansByBookID(bookID int) ([]*model.Loan, error) {
 func (s *Store) FindLoansByUserID(userID int) ([]*model.Loan, error) {
 	return nil, nil
 }
+
+func (s *Store) CreateLoan(loan *model.Loan) error {
+	args := s.Called(loan)
+	if args.Get(0) != nil {
+		return args.Error(0)
+	}
```
sfqi: add an empty line after the }

interactor/mock/loans.go
```diff
@@ -24,3 +24,11 @@ func (l *Loans) FindAll() ([]*model.Loan, error) {
 	}
 	return nil, args.Error(1)
 }
+
+func (l *Loans) CreateLoan(userId int, bookId int, state model.LoanType) error {
+	args := l.Called(userId, bookId, state)
+	if args.Get(0) != nil {
+		return args.Error(0)
+	}
```
sfqi: add an empty line after the }

interactor/loans.go
```diff
@@ -30,3 +37,12 @@ func (l *Loan) FindAll() ([]*model.Loan, error) {
 
 	return l.loansStore.FindAllLoans()
 }
+
+func (l *Loan) CreateLoan(userID int, bookID int, loanType model.LoanType) error {
```
sfqi: this exported `model.LoanType` was bugging me and I think I know why:
1. we're not using `NewBorrowedLoan` & `NewReturnedLoan`
2. to take advatage of 1. and to avoid spreading knowledge about the `LoanType` type, there should be two methods for creating a loan
3. maybe we should have two separate interactors: `LoanIndex` and `BookLoan`. Then `BookLoan` would have `Borrow` and `Return` as methods. Or maybe the `Borrow`/`Return` methods could belong to the `Book` interactor.

In this PR: do just 1. & 2. and we'll figure out 3. later

bokimilinkovic: @sfqi  do you think we should first check value for` loanType `inside `CreateLoan,` then call the proper constructor ( NewBorrowedLoan or NewReturnedLoan) ?

domain/model/loan.go
```diff
@@ -21,23 +20,19 @@ type Loan struct {
 	CreatedAt     time.Time
 }
 
-func NewLoan(userId int, bookId int, state loanType) (*Loan, error) {
-	uuid, err := uuid.NewUUID()
-	if err != nil {
-		return nil, err
-	}
+func NewLoan(userId int, bookId int, uuid string, loanType loanType) (*Loan, error) {
```
sfqi: We don't  need the error anymore

And maybe we could unexport `NewLoan`

interactor/loans.go
```diff
@@ -30,3 +37,21 @@ func (l *Loan) FindAll() ([]*model.Loan, error) {
 
 	return l.loansStore.FindAllLoans()
 }
+
+func (l *Loan) Borrow(userID int, bookID int) error {
+	uuid, err := l.generator.Do()
+	if err != nil {
+		return err
+	}
```
sfqi: Add an empty line after the `}`

interactor/loans.go
```diff
@@ -30,3 +37,21 @@ func (l *Loan) FindAll() ([]*model.Loan, error) {
...
+	loan, err := model.BorrowedLoan(userID, bookID, uuid)
+	return l.loansStore.CreateLoan(loan)
+}
+
+func (l *Loan) Return(userID int, bookID int) error {
+	uuid, err := l.generator.Do()
+	if err != nil {
+		return err
+	}
```
sfqi: Add an empty line after the `}`

interactor/loans_test.go
```diff
@@ -123,6 +124,73 @@ func TestFindAllLoans(t *testing.T) {
...
+
+		loan, err := model.BorrowedLoan(1, 1, "gen123-gen321")
+
+		generator.On("Do").Return("gen123-gen321", nil)
+
+		store.On("CreateLoan", loan).Return(nil)
+		l := interactor.NewLoan(store, generator)
+		err = l.Borrow(loan.UserID, loan.BookID)
+		assert.NoError(err)
+
```
sfqi: remove the empty line

service/uuid_test.go
```diff
@@ -0,0 +1,22 @@
...
+		t.Run(fmt.Sprintf("test %d", i), func(t *testing.T) {
+			g := &Generator{}
+			got, err := g.Do()
+			if err != nil {
+				t.Errorf("Do() error = %v", err)
+				return
+			}
+			if len(got) != 36 {
+				t.Errorf("uuid is not allowed lenght")
+			}
```
sfqi: I mean, we're using assert & require everywhere else, so might as well use it here

service/uuid_test.go
```diff
@@ -0,0 +1,22 @@
+package service
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestGenerator_Do(t *testing.T) {
+	for i := 0; i < 10; i++ {
```
sfqi: we probably don't need to iterate, especially since the test itself doesn't have a state

service/uuid_test.go
```diff
@@ -1,22 +1,20 @@
...
-			}
-		})
-	}
+	require := require.New(t)
+	t.Run(("test"), func(t *testing.T) {
+		g := &Generator{}
+		got, err := g.Do()
+		if err != nil {
+			t.Errorf("Do() error = %v", err)
+			return
```
sfqi: this should be a require

service/uuid_test.go
```diff
@@ -1,22 +1,20 @@
...
+		}
+		require.Equal(36, len(got))
```
sfqi: this should be an assert

PR #88, opened 2020-01-17
Added loan index method with tests in loanhandler

StefanGajic:
https://trello.com/c/jIhBpckp/124-add-loan-index-with-tests

### DIFF
--- a/handler/loanhandler.go
+++ b/handler/loanhandler.go
@@ -20,14 +20,14 @@ type loanInteractor interface {
 
 func (l *LoanHandler) Index(w http.ResponseWriter, r *http.Request) *HTTPError {
 	w.Header().Set("Content-Type", "application/json")
-	allLoans, err := l.Interactor.FindAll()
+	loans, err := l.Interactor.FindAll()
 	if err != nil {
 		return newHTTPError(http.StatusInternalServerError, err)
 	}
 
-	err = json.NewEncoder(w).Encode(allLoans)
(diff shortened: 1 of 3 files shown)


handler/loanhandler.go
```diff
@@ -18,5 +19,15 @@ type loanInteractor interface {
...
 func (l *LoanHandler) Index(w http.ResponseWriter, r *http.Request) *HTTPError {
+	w.Header().Set("Content-Type", "application/json")
+	allLoans, err := l.Interactor.FindAll()
+	if err != nil {
+		return newHTTPError(http.StatusInternalServerError, err)
+	}
+
+	err = json.NewEncoder(w).Encode(allLoans)
+	if err != nil {
+		return newHTTPError(http.StatusBadRequest, err)
```
sfqi: Not bad request, internal server error

handler/loanhandler.go
```diff
@@ -18,5 +19,15 @@ type loanInteractor interface {
 }
 
 func (l *LoanHandler) Index(w http.ResponseWriter, r *http.Request) *HTTPError {
+	w.Header().Set("Content-Type", "application/json")
+	allLoans, err := l.Interactor.FindAll()
```
sfqi: I mean, you can just call it `loans`

handler/loanhandler.go
sfqi: And if this is copied from somewhere,go ahead and fix it there as well

PR #89, opened 2020-01-17
Feature/books/id/loans endpoint

bokimilinkovic:
https://trello.com/c/1ray6POt/125-add-books-bookid-loans-endpoint-with-tests

### DIFF
--- a/handler/loanhandler.go
+++ b/handler/loanhandler.go
@@ -1,9 +1,14 @@
 package handler
 
 import (
-	"net/http"
-
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/gorilla/mux"
 	"github.com/sfqi/library/domain/model"
+	"github.com/sfqi/library/handler/dto"
+	"net/http"
+	"strconv"
 )
 
 type LoanHandler struct {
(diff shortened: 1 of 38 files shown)


seed/main.go
```diff
@@ -32,6 +32,30 @@ var books = []*model.Book{
...
+	{
+		ID:            2,
+		TransactionID: "xzxc2cxcc",
+		UserID:        2,
+		BookID:        1,
+		Type:          0,
+	},
+	{
+		ID:            3,
+		TransactionID: "cccscscscs2131",
```
sfqi: we should use uuids in the seed

handler/loanhandler.go
```diff
@@ -1,22 +1,50 @@
...
 	FindAll() ([]*model.Loan, error)
 	FindByUserID(id int) ([]*model.Loan, error)
 	FindByBookID(id int) ([]*model.Loan, error)
 }
 
-func (l *LoanHandler) Index(w http.ResponseWriter, r *http.Request) *HTTPError {
+func (l *LoanHandler) FindLoansByBookID(w http.ResponseWriter, r *http.Request) *HTTPError {
+	id, err := strconv.Atoi(mux.Vars(r)["book_id"])
+	if err != nil {
+		return newHTTPError(http.StatusBadRequest, err)
```
sfqi: I'm kinda leaning towards making this a not found error

handler/loanhandler.go
```diff
@@ -1,22 +1,50 @@
...
+	}
+	loans, err := l.Interactor.FindByBookID(id)
+	if err != nil {
+		return newHTTPError(http.StatusInternalServerError, err)
+	}
+	if len(loans) == 0 {
+		return newHTTPError(http.StatusNotFound, errors.New("No loans for given book id can be found"))
```
sfqi: maybe this shouldn't be an error, since it's possible that a book just hasn't been loaned yet
how about we just return an empty json array - I think that approach is more convenient for the frontend

handler/loanhandler.go
```diff
@@ -1,22 +1,50 @@
...
+	}
+	var loanResponses []*dto.LoanResponse
+	for _, loan := range loans {
+		loanResponses = append(loanResponses, toLoanResponse(*loan))
+	}
+	err = json.NewEncoder(w).Encode(loanResponses)
+	if err != nil {
+		return newHTTPError(http.StatusBadRequest, err)
```
sfqi: internal server error

handler/loanhandler.go
```diff
@@ -1,22 +1,50 @@
...
+		return newHTTPError(http.StatusNotFound, err)
+	}
+
+	loans, err := l.Interactor.FindByBookID(id)
+	if err != nil {
+		return newHTTPError(http.StatusInternalServerError, err)
+	}
+	if len(loans) == 0 {
+		err = json.NewEncoder(w).Encode(loans)
+	}
```
sfqi: The error isn't handled. And we don't want to continue executing if the response has already been encoded.

But also: do we have to check whether the array is empty? What happens if the endpoint just encodes the array, regardless of how many elements it has?

PR #91, opened 2020-01-20
Used transactions in creating loans and books in seeds

bokimilinkovic:
https://trello.com/c/zEZ3k3fa/130-use-a-transaction-to-rollback-seeding-if-an-error-occurs

### DIFF
--- a/repository/postgres/postgres.go
+++ b/repository/postgres/postgres.go
@@ -24,7 +24,7 @@ func NewStore(db *gorm.DB) *Store {
 	return &Store{db: db}
 }
 
-func (s *Store) GetDB() *gorm.DB {
+func (s *Store) DB() *gorm.DB {
 	return s.db
 }
 
--- a/repository/postgres/postgres.go
+++ b/repository/postgres/postgres.go
@@ -24,6 +24,10 @@ func NewStore(db *gorm.DB) *Store {
 	return &Store{db: db}
 }
 
+func (s *Store) GetDB() *gorm.DB {
+	return s.db
+}
+
 func Open(config PostgresConfig) (*Store, error) {
 	psqlInfo := fmt.Sprintf("host=%s port=%d user=%s password=%s dbname=%s sslmode=disable",
 		config.Host, config.Port, config.User, config.Password, config.Name)
--- a/seed/main.go
+++ b/seed/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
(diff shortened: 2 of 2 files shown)


seed/main.go
```diff
@@ -47,10 +48,50 @@ func main() {
...
+	}()
+
+	if err := tx.Error; err != nil {
+		return err
+	}
 	for _, book := range books {
-		if err := store.CreateBook(book); err != nil {
-			panic(err)
+		if err := tx.Create(book).Error; err != nil {
+			tx.Rollback()
```
sfqi: you're rollbacking twice this way, once here, and once when executing the recovery defer

seed/main.go
```diff
@@ -47,10 +48,50 @@ func main() {
...
+	if err := tx.Error; err != nil {
+		return err
+	}
+	for _, loan := range loans {
+		if err := tx.Create(loan).Error; err != nil {
+			tx.Rollback()
+			return err
+		}
+	}
+	return tx.Commit().Error
```
sfqi: We should commit at the end of the seed. It doesn't really make sense to have multiple commits, since you'll end up making only partial rollbacks if there's an error

seed/main.go
```diff
@@ -47,10 +48,50 @@ func main() {
...
+
+	err = CreateBooks(store.GetDB(), books)
+	if err != nil {
+		panic(err)
+	}
+
+}
+
+func CreateBooks(db *gorm.DB, books []*model.Book) error {
+	tx := db.Begin()
```
sfqi: the transaction should begin at the begging of the seed, and there should be only one

seed/main.go
```diff
@@ -47,10 +48,50 @@ func main() {
...
+	defer func() {
+		if r := recover(); r != nil {
+			tx.Rollback()
+		}
+	}()
```
sfqi: this should be in main

repository/postgres/postgres.go
```diff
@@ -24,6 +24,10 @@ func NewStore(db *gorm.DB) *Store {
 	return &Store{db: db}
 }
 
+func (s *Store) GetDB() *gorm.DB {
```
sfqi: Go discourages methods starting with `Set` and `Get`. Can we call it `DB()`?

sfqi:
@bokimilinkovic can you do loans as well?

PR #92, opened 2020-01-20
Check if tables are empty before seeding

StefanGajic:
https://trello.com/c/3MkKDlN7/129-check-whether-all-tables-are-empty-before-seeding

### DIFF
--- a/seed/main.go
+++ b/seed/main.go
@@ -51,19 +51,25 @@ var loans = []*model.Loan{
 }
 
 func areTablesEmpty(db *gorm.DB) error {
-	var bookNum int
-	var loanNum int
-	if err := db.Model(&model.Book{}).Count(&bookNum).Error; err != nil {
-		return err
-	}
-	if err := db.Model(&model.Loan{}).Count(&loanNum).Error; err != nil {
-		return err
+	var bookCount int
+	var loanCount int
+	var count int
+
+	switch count {
+	case bookCount:
+		if err := db.Model(&model.Book{}).Count(&bookCount).Error; err != nil {
+			return err
+		}
+	case loanCount:
+		if err := db.Model(&model.Loan{}).Count(&loanCount).Error; err != nil {
+			return err
+		}
 	}
 
-	if bookNum != 0 {
+	switch count {
+	case bookCount:
 		return errors.New("Table book is not empty!")
-	}
-	if loanNum != 0 {
+	case loanCount:
 		return errors.New("Table loan is not empty!")
 	}
 	return nil
--- a/seed/main.go
+++ b/seed/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/sfqi/library/domain/model"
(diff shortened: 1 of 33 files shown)


seed/main.go
```diff
@@ -47,10 +48,19 @@ func main() {
...
-		if err := store.CreateBook(book); err != nil {
-			panic(err)
+
+	db := store.GetDB()
+
+	bookList := []model.Book{}
+	var num int
+	db.Model(&bookList).Count(&num)
+
+	if num < 0 {
```
bokimilinkovic: ==

seed/main.go
```diff
@@ -47,10 +48,19 @@ func main() {
...
-		if err := store.CreateBook(book); err != nil {
-			panic(err)
+
+	db := store.GetDB()
+
+	bookList := []model.Book{}
+	var num int
+	db.Model(&bookList).Count(&num)
+
+	if num == 0 {
```
sfqi: Use an early return (`if num != 0`)

sfqi: Also `num` isn't great as a name. Maybe write a function that will check whether the table is empty - for readability

seed/main.go
```diff
@@ -32,6 +34,16 @@ var books = []*model.Book{
 	},
 }
 
+func areTablesEmpty(db *gorm.DB) bool {
+	var num int
+	db.Model([]*model.Book{}).Count(&num)
+
+	if num > 1 {
+		return true
```
sfqi: this can't be right

seed/main.go
```diff
@@ -47,10 +59,13 @@ func main() {
...
+
+	if !areTablesEmpty(store.GetDB()) {
+		for _, book := range books {
+			if err := store.CreateBook(book); err != nil {
+				panic(err)
+			}
 		}
 	}
-
+	fmt.Println("Tables are not empty")
```
sfqi: This doesn't really make sense. We should panic if the tables are empty and that's it

seed/main.go
```diff
@@ -47,10 +58,13 @@ func main() {
...
+
+	if !areTablesEmpty(store.GetDB()) {
+		for _, book := range books {
+			if err := store.CreateBook(book); err != nil {
+				panic(err)
+			}
 		}
 	}
-
+	panic("Tables are not empty")
```
sfqi: It would be nice to know which table isn't empty

seed/main.go
sfqi: What's the status on this?

StefanGajic: I have changed it to 	```
if num > 0 {
		return true
```
I am not sure if you saw it

Comment (sfqi) [reply seed/main.go]: So what this function says it that if there's more than 0 rows in the table, the table is empty? Am I misreading anything?

File: seed/main.go [hunk 6]
```diff
@@ -32,6 +33,16 @@ var books = []*model.Book{
 	},
 }
 
+func areTablesEmpty(db *gorm.DB) bool {
+	var num int
+	db.Model([]*model.Book{}).Count(&num)
```
sfqi: looking at this example from gorm:
```
db.Model(&User{}).Where("name = ?", "jinzhu").Count(&count)
```

I'm guessing we don't need a slice here

sfqi:
Also, could you merge https://github.com/sfqi/library/pull/91 in here and do the same for loans?

seed/main.go
```diff
@@ -1,37 +1,74 @@
...
+		return err
+	}
+
+	if bookNum != 0 {
+		return errors.New("Table book is not empty!")
+	}
+	if loanNum != 0 {
+		return errors.New("Table loan is not empty!")
+	}
+	return nil
```
sfqi: you might be noticing that the code is a bit repetitive here (imagine we had 12 tables)
can you refactor it?

also, why `num`? `count`!

sfqi: 1. You  declare `count`, it's automatically set to zero, and then you switch based on the value of count? It's always zero. (so are `bookCount` and `loanCount`) 
2. You didn't really achieve much in the way of having less repetition. Again, imagine there are 12 tables - you'd end up with a massive function
3. Hint: you need only one `count` var

PR #93, opened 2020-01-20
Added public message field to HTTPError and return instead of Error()

StefanGajic:
https://trello.com/c/cDdDqBlu/131-add-a-public-message-field-to-http-error-type

### DIFF
--- a/handler/bookhandler.go
+++ b/handler/bookhandler.go
@@ -46,7 +46,7 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) *HTTPError {
 	w.Header().Set("Content-Type", "application/json")
 	allBooks, err := b.Interactor.FindAll()
 	if err != nil {
-		return newHTTPError(http.StatusInternalServerError, err, "error find all books")
+		return newHTTPError(http.StatusInternalServerError, err)
 	}
 	var bookResponses []*dto.BookResponse
 
@@ -57,7 +57,7 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) *HTTPError {
 
 	err = json.NewEncoder(w).Encode(bookResponses)
 	if err != nil {
-		return newHTTPError(http.StatusInternalServerError, err, "error index encoding")
+		return newHTTPError(http.StatusInternalServerError, err)
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


handler/bookhandler.go
```diff
@@ -46,7 +46,7 @@ func (b *BookHandler) Index(w http.ResponseWriter, r *http.Request) *HTTPError {
 	w.Header().Set("Content-Type", "application/json")
 	allBooks, err := b.Interactor.FindAll()
 	if err != nil {
-		return newHTTPError(http.StatusInternalServerError, err)
+		return newHTTPError(http.StatusInternalServerError, err, "error find all books")
```
sfqi: 1. No need to add these messages everywhere. We'll add them later - where needed.
2. I want us to be able to do something like `newHTTPError(<errorCode>, err).PublicMessage("public message")` - the same approach as with context

handler/error.go
```diff
@@ -10,22 +10,24 @@ import (
...
+	code      int
+	internal  error
+	context   string
+	publicMsg string
 }
 
 func (h HTTPError) Error() string {
 	if h.context != "" {
-		return fmt.Sprintf("HTTP %d: %s: %s", h.code, h.context, h.internal)
+		return fmt.Sprintf("HTTP %d: %s: %s: %s", h.code, h.context, h.internal, h.publicMsg)
```
sfqi: No need to add the public message to `Error`'s output

handler/error.go
```diff
@@ -10,22 +10,24 @@ import (
...
 	}
-	return fmt.Sprintf("HTTP %d: %s", h.code, h.internal)
+	return fmt.Sprintf("HTTP %d: %s: %s", h.code, h.internal, h.publicMsg)
 }
 
-func newHTTPError(code int, err error) *HTTPError {
+func newHTTPError(code int, err error, msg string) *HTTPError {
```
sfqi: No need to add new arguments here

handler/error.go
```diff
@@ -51,7 +53,7 @@ func (eh ErrorHandler) Wrap(handler customHandler) http.Handler {
 			eh.Logger.Error(err)
 
 			if err.code < http.StatusInternalServerError {
-				errorMsg = err.Error()
+				errorMsg = err.publicMsg
```
sfqi: I'd prefer to have a method (`publicError`, or a better name, if you come up with it) which returns the `publicMsg` field if it's not empty, and returns `Error()` otherwise

PR #95, opened 2020-01-22
Feature/books/id/borrow&return endpoints

StefanGajic:
We merged 2 branches together so we would avoid future conflicts and simplified it 

https://trello.com/c/3USQwytv/133-add-books-bookid-borrow-endpoint

https://trello.com/c/lbiToQRy/134-add-books-bookid-return-endpoint

### DIFF
--- a/handler/dto/loan.go
+++ b/handler/dto/loan.go
@@ -1,5 +1,7 @@
 package dto
 
+import "github.com/sfqi/library/domain/model"
+
 type LoanResponse struct {
 	ID            int    `json:"id"`
 	TransactionID string `json:"transaction_id"`
@@ -8,12 +10,12 @@ type LoanResponse struct {
 	Type          string `json:"loan_type"`
 }
 
-func CreateLoanResponse(id int, transactionID string, userID int, bookID int, loanType string) (*LoanResponse, error) {
+func CreateLoanResponse(l *model.Loan, loanType string) *LoanResponse {
 	return &LoanResponse{
-		ID:            id,
-		TransactionID: transactionID,
-		UserID:        userID,
-		BookID:        bookID,
+		ID:            l.ID,
+		TransactionID: l.TransactionID,
+		UserID:        l.UserID,
+		BookID:        l.BookID,
 		Type:          loanType,
-	}, nil
+	}
 }
--- a/handler/dto/loan.go
+++ b/handler/dto/loan.go
@@ -1,21 +1,9 @@
 package dto
 
-import "github.com/sfqi/library/domain/model"
-
 type LoanResponse struct {
 	ID            int    `json:"id"`
 	TransactionID string `json:"transaction_id"`
 	UserID        int    `json:"user_id"`
 	BookID        int    `json:"book_id"`
 	Type          string `json:"loan_type"`
-}
-
-func CreateLoanResponse(l *model.Loan, loanType string) *LoanResponse {
-	return &LoanResponse{
-		ID:            l.ID,
-		TransactionID: l.TransactionID,
-		UserID:        l.UserID,
-		BookID:        l.BookID,
-		Type:          loanType,
-	}
 }
--- a/handler/dto/loan.go
+++ b/handler/dto/loan.go
@@ -8,12 +8,12 @@ type LoanResponse struct {
 	Type          string `json:"loan_type"`
 }
 
-func CreateLoanResponse(id int, transactionID string, userID int, bookID int, loanType string) *LoanResponse {
+func CreateLoanResponse(id int, transactionID string, userID int, bookID int, loanType string) (*LoanResponse, error) {
 	return &LoanResponse{
 		ID:            id,
 		TransactionID: transactionID,
 		UserID:        userID,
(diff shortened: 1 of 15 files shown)


handler/dto/loan.go
```diff
@@ -8,12 +8,12 @@ type LoanResponse struct {
...
-func CreateLoanResponse(id int, transactionID string, userID int, bookID int, loanType string) *LoanResponse {
+func CreateLoanResponse(id int, transactionID string, userID int, bookID int, loanType string) (*LoanResponse, error) {
 	return &LoanResponse{
 		ID:            id,
 		TransactionID: transactionID,
 		UserID:        userID,
 		BookID:        bookID,
 		Type:          loanType,
-	}
+	}, nil
```
sfqi: do we need the error here?

handler/loanhandler.go
```diff
@@ -43,6 +56,42 @@ func (l *LoanHandler) FindLoansByBookID(w http.ResponseWriter, r *http.Request)
...
+		return newHTTPError(http.StatusInternalServerError, err)
+	}
+	w.Write([]byte("Loan successfully createad"))
+	return nil
+}
+
+func toLoanResponse(l *model.Loan) (*dto.LoanResponse, *HTTPError) {
+	loanType, err := l.PrintType()
+	if err != nil {
+		return nil, newHTTPError(http.StatusNotFound, err)
```
sfqi: why not found?

handler/loanhandler.go
```diff
@@ -43,6 +56,42 @@ func (l *LoanHandler) FindLoansByBookID(w http.ResponseWriter, r *http.Request)
...
+	}
+
+	loanResponse, err := dto.CreateLoanResponse(l.ID, l.TransactionID, l.UserID, l.BookID, loanType)
```
sfqi: could we just pass the loan model to this func (like in `toBookResponse`)?
also: we should have the same approach to creating responses for every model, whatever the approach is a func in `dto` or in a handler file

interactor/borrow_return.go
```diff
@@ -0,0 +1,43 @@
+package interactor
+
+import "github.com/sfqi/library/domain/model"
+
+type borrowReturn interface {
```
sfqi: I don't like this name :)

interactor/borrow_return.go
```diff
@@ -0,0 +1,43 @@
...
+	CreateLoan(*model.Loan) error
+}
+
+type BookLoan struct {
```
sfqi: What if we had `LoanReader` (with get & index methods) and `LoanWriter` (for create methods)?

handler/loanhandler.go
```diff
@@ -43,6 +56,43 @@ func (l *LoanHandler) FindLoansByBookID(w http.ResponseWriter, r *http.Request)
...
 }
 
-func toLoanResponse(b model.Loan) *dto.LoanResponse {
-	return dto.CreateLoanResponse(b.ID, b.TransactionID, b.UserID, b.BookID, b.PrintType())
+func (n *NewLoanHandler) BorrowBook(w http.ResponseWriter, r *http.Request) *HTTPError {
+	id, err := strconv.Atoi(mux.Vars(r)["id"])
+	if err != nil {
+		return newHTTPError(http.StatusNotFound, err)
+	}
+	err = n.Interactor.Borrow(10, id)
```
sfqi: what's `10`?
the user id should be submitted in the request

main.go
```diff
@@ -78,6 +81,9 @@ func main() {
 
 	//loans endpoints
 	s.Handle("/{id}/loans", handleFunc(loanHandler.FindLoansByBookID)).Methods("GET")
+	s.Handle("/{id}/borrow", handleFunc(newLoanHandler.BorrowBook)).Methods("GET")
```
sfqi: POST!

main.go
```diff
@@ -78,6 +81,9 @@ func main() {
...
+
+	s.Handle("/{id}/return", handleFunc(newLoanHandler.ReturnBook)).Methods("GET")
```
sfqi: POST!

handler/loanhandler.go
```diff
@@ -10,16 +10,25 @@ import (
...
 
-type loanInteractor interface {
+type loanReader interface {
 	FindByID(ID int) (*model.Loan, error)
 	FindAll() ([]*model.Loan, error)
 	FindByUserID(id int) ([]*model.Loan, error)
 	FindByBookID(id int) ([]*model.Loan, error)
 }
 
+type NewLoanHandler struct {
```
sfqi: `NewLoanHandler`? Come on

Since we have separate interactors for reading and writing loans, let's do the same for handlers. How about `ReadLoanHandler` and `WriteLoanHandler`?

handler/loanhandler.go
bokimilinkovic: it's temporarily there, because all user operations are on another branch. I am guessing we will also get userID by url params.

sfqi: mmm, leave it as is at the moment

PR #97, opened 2020-01-29
Changing field 'availale' depending on endpoint

bokimilinkovic:
https://trello.com/c/TscF6Er6/135-set-the-available-field-to-true-false-when-book-is-returned-borrowed

### DIFF
--- a/handler/loanhandler.go
+++ b/handler/loanhandler.go
@@ -2,6 +2,7 @@ package handler
 
 import (
 	"encoding/json"
+	"errors"
 	"net/http"
 	"strconv"
 
@@ -26,8 +27,8 @@ type WriteLoanHandler struct {
 }
 
 type loanWriter interface {
-	Borrow(userID int, bookID int) error
-	Return(userID int, bookID int) error
+	Borrow(userID int, book *model.Book) error
+	Return(userID int, book *model.Book) error
 }
 
 func (rl *ReadLoanHandler) Index(w http.ResponseWriter, r *http.Request) *HTTPError {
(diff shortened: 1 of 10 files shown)


interactor/borrow_return.go
```diff
@@ -28,6 +33,15 @@ func (l *LoanWriter) Borrow(userID int, bookID int) error {
 		return err
 	}
 
+	book, err := l.store.FindBookById(bookID)
```
sfqi: Don't we have a book loader middleware?

bokimilinkovic: yees, i forgot that we have book in context

bokimilinkovic: but if I get book from context, in loanhandler, I need to call UpdateBook directly there, or I can pass somehow that book in  loanWriter Interactor and than from there to call UpdateBook ? Which one is better ? @sfqi

sfqi: you pass the book to the interactor

handler/loanhandler.go
```diff
@@ -114,11 +115,11 @@ func (wl *WriteLoanHandler) BorrowBook(w http.ResponseWriter, r *http.Request) *
 }
 
 func (wl *WriteLoanHandler) ReturnBook(w http.ResponseWriter, r *http.Request) *HTTPError {
-	id, err := strconv.Atoi(mux.Vars(r)["id"])
-	if err != nil {
-		return newHTTPError(http.StatusNotFound, err)
+	book := r.Context().Value("book")
```
sfqi: you need to check the `ok` as well, otherwise it might panic

sfqi:
@bokimilinkovic See if you can use git revert to undo this commit: https://github.com/sfqi/library/pull/97/commits/2afe6e17ccc20a9dd9e23038b480209b0610dc50

We won't be using the book from context

PR #99, opened 2020-02-04
All tests in handler are passing

StefanGajic:
https://trello.com/c/lmBQdE1S/137-fix-tests-in-handlers-which-are-not-passing

### DIFF
--- a/handler/bookhandler_test.go
+++ b/handler/bookhandler_test.go
@@ -105,7 +105,7 @@ func TestIndex(t *testing.T) {
 		handler := bookHandler.Index
 		httperror := handler(rr, req)
 
-		expectedResponse := "HTTP 500: Error finding books"
+		expectedResponse := "HTTP 500: Error finding books: "
 		assert.Equal(expectedResponse, httperror.Error())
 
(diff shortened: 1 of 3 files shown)


handler/bookhandler_test.go
```diff
@@ -105,7 +105,7 @@ func TestIndex(t *testing.T) {
 		handler := bookHandler.Index
 		httperror := handler(rr, req)
 
-		expectedResponse := "HTTP 500: Error finding books"
+		expectedResponse := "HTTP 500: Error finding books: "
```
sfqi: why is this happening now?
why do all errors end with `: `?

StefanGajic: I think it is because we added `:` 
I tried to remove colon but actual needs `:`  at the end

bokimilinkovic: this should be optional, if we have some additional context for error message

sfqi: If every error message ends with `: `, we should fix that. As in, they shouldn't all end with `: `

sfqi: Let's close this PR and deal with the issue properly

PR #100, opened 2020-02-04
Read lock used

bokimilinkovic:
What are we doing here, is adding one more method against postgres 'FindBookByIDForUpdate' . We are doing this so we can avoid non-consistent reading from data base. If we get 2 request for borrowing book at a same time, both of them can pass, and we want to avoid that.
If we have 2 threads hitting the same endpoint, if 1 thread has a read lock on a row/table, no thread can update/insert/delete data from that table. The write transaction should wait for read locks to finish reading.

### DIFF
--- a/interactor/store.go
+++ b/interactor/store.go
@@ -16,4 +16,5 @@ type Store interface {
 	Transaction() Store
 	Commit() error
 	Rollback()
+	FindByIdSelectUpdate(int) (*model.Book, error)
 }
--- a/interactor/store.go
+++ b/interactor/store.go
@@ -16,5 +16,5 @@ type Store interface {
 	Transaction() Store
 	Commit() error
 	Rollback()
-	FindByIdSelectUpdate(int) (*model.Book, error)
+	FindBookByIDForUpdate(int) (*model.Book, error)
 }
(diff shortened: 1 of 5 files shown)


interactor/store.go
```diff
@@ -16,4 +16,5 @@ type Store interface {
 	Transaction() Store
 	Commit() error
 	Rollback()
+	FindByIdSelectUpdate(int) (*model.Book, error)
```
sfqi: `FindBookByIDForUpdate`

repository/postgres/postgres.go
```diff
@@ -134,3 +134,11 @@ func (store *Store) Commit() error {
 func (store *Store) Rollback() {
 	store.db.Rollback()
 }
+
+func (store *Store) FindByIdSelectUpdate(bookID int) (*model.Book, error) {
+	var book model.Book
+	if err := store.db.Set("gorm:query_option", "FOR UPDATE").First(&model.Book{}, bookID).Scan(&book).Error; err != nil {
```
sfqi: Can't we do `.First(&book, bookID).Error`?

`bookID` -> 'id'

sfqi:
@bokimilinkovic Could you write a description of the PR so that @StefanGajic & @zzokki81 can have a better idea of what is being done - and why it's being done?

PR #106, opened 2020-02-07
Replaced t.Fatal with require

zzokki81:
https://trello.com/c/sow9PoLY/143-replace-tfatal-with-require

### DIFF
--- a/handler/bookhandler_test.go
+++ b/handler/bookhandler_test.go
@@ -52,7 +52,7 @@ func TestIndex(t *testing.T) {
 
 		req, err := http.NewRequest("GET", "/books", nil)
 		if err != nil {
-			t.Fatal(err)
+			require.NoError(err)
 		}
 
 		rr := httptest.NewRecorder()
@@ -95,7 +95,7 @@ func TestIndex(t *testing.T) {
 
 		req, err := http.NewRequest("GET", "/books", nil)
 		if err != nil {
-			t.Fatal(err)
+			require.NoError(err)
 		}
 
 		rr := httptest.NewRecorder()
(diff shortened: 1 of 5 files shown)


handler/bookhandler_test.go
```diff
@@ -52,7 +52,7 @@ func TestIndex(t *testing.T) {
 
 		req, err := http.NewRequest("GET", "/books", nil)
 		if err != nil {
-			t.Fatal(err)
+			require.NoError(err)
 		}
```
sfqi: you don't need to check whether the error is nil in order to use `require.NoError`

middleware/bookloader_test.go
```diff
@@ -50,31 +53,29 @@ func TestGetBook(t *testing.T) {
 		req, err := http.NewRequest("GET", "/{id}", nil)
 		params := map[string]string{"id": "6"}
 		req = mux.SetURLVars(req, params)
-		if err != nil {
-			t.Fatal(err)
-		}
+
+		require.NoError(err)
```
sfqi: move this two lines up (right after `req, err := http.NewRequest("GET", "/{id}", nil)`)

middleware/bookloader_test.go
```diff
@@ -50,31 +53,29 @@ func TestGetBook(t *testing.T) {
...
-		bookLoader := middleware.BookLoader{}
+		bookLoader := middleware.BookLoader{Logger: log.New()}
 
 		req, err := http.NewRequest("GET", "/{id}", nil)
 		params := map[string]string{"id": "1"}
 		req = mux.SetURLVars(req, params)
-		if err != nil {
-			t.Fatal(err)
-		}
+		require.NoError(err)
```
sfqi: two lines up

PR #107, opened 2020-02-14
Fix seeding for users

zzokki81:


sfqi:
I'm a bit confused, what was the motivation and reasoning behind those changes?

zzokki81:
> I'm a bit confused, what was the motivation and reasoning behind those changes?

Every time when we start main.go in seed, book and loan will put data in database.. but for user not, it always remain empty...

seed/main.go
```diff
@@ -56,41 +41,50 @@ var users = []*model.User{
...
 		Email:     "john@smith.com",
 		Name:      "John",
-		LastName:  "Smith",
+		Lastname:  "Smith",
 		CreatedAt: time.Time{},
 		UpdatedAt: time.Time{},
 	},
 }
 
-func areTablesEmpty(db *gorm.DB) error {
```
StefanGajic: why did you remove this?

seed/main.go
```diff
@@ -108,47 +102,21 @@ func main() {
 		panic(err)
 	}
 	defer store.Close()
-
-	tx := store.DB().Begin()
```
StefanGajic: why did you remove transactions?

seed/main.go
```diff
@@ -108,47 +102,21 @@ func main() {
...
-
-	defer func() {
-		store.Close()
-		if r := recover(); r != nil {
-			fmt.Println(r.(error))
-			tx.Rollback()
-		}
-	}()
-
-	if err = areTablesEmpty(store.DB()); err != nil {
```
StefanGajic: this is something we need to check if tables are empty

